PR #2, opened 2018-04-13
Istio-OPA Plugin

ashutosh-narkar:
Istio plugin that implements the _Check_ api and uses _OPA_ to make policy decisions


tsandall:
This is great! Here are a bunch comments from my first look at the project. I'm leaving this at the PR level because the change set is big enough to make the diff laggy.

README:

* The indention in the app deployment section needs to be fixed

Plugin config:

* Rename the sample_config.json to sample_config.yaml
* Use underscore in the plugin name instead of hyphen for consistency
* The plugin configuration needs to be nested under `plugins`

Build:

* `make image` should tag images with a point release to differentiate plugin versions, right now it's just tagged as `<opa-version>-istio` we need some kind of version identifier for the plugin. I think that an extra semver is overkill. Perhaps just an integer.

* `make image` fails because Docker cannot find `config_istio.json`.

* OPA version info is not set. The build will need to be fixed to set the version correctly (e.g., run `./opa_istio_darwin_amd64 version` to reproduce.)

Logging:

* OPA tries to provide consistent logging. This is a good guide for the content: https://nedbatchelder.com/text/log-style-guide.html. Let's try to follow this in the official plugins.

Examples:

* The config/demo/bookinfo.yaml file refers to openpolicyagent/istio-proxy:0c3a (which it should not IIUC). Also, the OPA container config appears to be incorrect (missing --server, bad image name.) See similar note about docs below.

* Consider renaming `config` directory to `sample` or something that indicates these files are for test/example purposes (i.e., they're not required configuration.)

Docs:

* Under the "Install Istio" section, it would be useful to provide a bit more detail on the mesh config update. I had installed Istio from scratch w/o using the Istio install config from this repo. I'm not sure if Istio supports modifications to the mesh config. We should test this and include a note about whether we can modify this config or if the plugin requires a reinstall of Istio.

* Under the "Install BookInfo App" section, it would be better if the YAML snippet had the actual Docker image name instead of `<plugin-image>`. Also, use `--config-file` instead of `-c` for consistency. Also, `--server` is required.

* Code blocks in markdown should be given syntax highlighting, e.g., "\`\`\`yaml" or "```bash" and so on.

Misc.

* Consider using Internal packages for this code. I don't think we expect people to depend on it. https://stackoverflow.com/questions/33351387/how-to-use-internal-packages 

* Missing license notice from Istio attribute code.

* Mention that we need to attribute#GlobaList if upstream Istio changes.

tsandall:
Once you're ready, please squash the commits. At minimum, use two commits (one for vendored code, one for everything else.)

PR #11, opened 2018-04-25
Update OPA version 

ashutosh-narkar:
Update the OPA version in README, glide.yaml and quick_start.yaml


build/update-opa-version.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 # Script to revendor OPA, Add and Commit changes if needed
 
+set +e
```
tsandall: Did you mean for this to be +e? +e will make the script NOT exit if subprocesses error.

PR #75, opened 2019-02-22
Refactor istio-opa into a custom plugin.

ashutosh-narkar:


internal/internal.go
```diff
@@ -148,34 +162,46 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
...
+		"decision":            result.decision,
 		"err":                 err,
-		"txn":                 txnID,
-		"metrics":             metrics.All(),
+		"txn":                 result.txnID,
+		"metrics":             result.metrics.All(),
 		"total_decision_time": time.Since(start),
 	}).Info("Returning policy decision.")
 
+	defer p.log(ctx, input, result, err)
```
tsandall: There is no reason to defer here.

docs/plugin.md
```diff
@@ -0,0 +1,55 @@
...
+
+    ```bash
+    make build-plugin
+    ```
+
+5. Create a configuration file.
+
+    ```yaml
+    decision_logs:
+        plugin: envoy.ext_authz.grpc
```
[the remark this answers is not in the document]
ashutosh-narkar: Removed.

main.go
```diff
@@ -27,8 +27,10 @@ func (Factory) Validate(m *plugins.Manager, config []byte) (interface{}, error)
...
 }
 
-func init() {
+// Init register plugin factory with OPA.
+func Init() error {
 	runtime.RegisterPlugin("envoy.ext_authz.grpc", Factory{})
+	return nil
 }
 
 func main() {
```
tsandall: It's a bit odd that we still have the `main` function since we're not changing the entry point behaviour at all. I'm guessing this is due to the fact the plugin build uses the vendored OPA which is incompatible with non-vendored OPA packages in built OPA executables? This is more of a question than a comment. Is it possible to get rid of the `main` function? If not, a comment explaining why we need it would be useful.

internal/internal.go
```diff
@@ -148,34 +162,49 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
...
+		"decision":            result.decision,
 		"err":                 err,
-		"txn":                 txnID,
-		"metrics":             metrics.All(),
+		"txn":                 result.txnID,
+		"metrics":             result.metrics.All(),
 		"total_decision_time": time.Since(start),
 	}).Info("Returning policy decision.")
 
+	err = p.log(ctx, input, result, err)
```
tsandall: If the decision logger returns an error, the logrus message above is going to be confusing. I think we ought to make them consistent.

main.go
ashutosh-narkar: Added a comment.

internal/internal.go
ashutosh-narkar: Done.

PR #93, opened 2019-06-20
Add pre processed path to the input.

ashutosh-narkar:
Add path as a string array to the input to help with indexing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@
 This repository contains an extended version of OPA (**OPA-Istio**) that allows you to enforce OPA
 policies at the Istio Proxy layer.
 
+## Issue Management
+Use [OPA GitHub Issues](https://github.com/open-policy-agent/opa/issues) to request features or file bugs.
+
 ## Overview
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -5,6 +5,9 @@
 This repository contains an extended version of OPA (**OPA-Istio**) that allows you to enforce OPA
 policies at the Istio Proxy layer.
 
+## Issue Management
+Use [OPA GitHub Issues](https://github.com/open-policy-agent/opa/issues) to request features or file bugs.
```
tsandall: Typically I'd do this in another commit :D

README.md
```diff
@@ -209,6 +212,7 @@ The `input` value defined for your policy will resemble the JSON below:
 
 ```json
 {
+  "parsed_path": ["api", "v1", "products"],
```
tsandall: It's worth documenting this field a little bit more since everything else in the input comes from the standard Envoy ext_authz check request type. I'd explicitly call this field out and mention when/how to use it.

ashutosh-narkar: Done

PR #96, opened 2019-07-08
Migrate to dep

ashutosh-narkar:
This pull request contains changes for switching from using `glide` as a package manager to `dep`. 

This PR features changes to use the Go implementation of Envoy's  `data-plane-api`.

Closes https://github.com/open-policy-agent/opa/issues/1512.


Gopkg.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[[constraint]]
+  name = "github.com/gogo/googleapis"
+  revision = "d31c731455cb061f42baff3bda55bad0118b126b"
+
+[[constraint]]
+  name = "google.golang.org/grpc"
+  version = "1.10.0"
+
+[[override]]
```
patrick-east: Should we add a comment for why we use the `override`?

ashutosh-narkar: Added a note about this.

PR #98, opened 2019-07-23
Add functionality to return more attributes of the Envoy Ext Authz re…

ashutosh-narkar:
…sponse.

The PR adds support for the `response_query` config field which allows the authorization service to return the status of request (ie. `allowed` or `denied`) alongwith some headers and body data to send back to the downstream client or upstream.

See: https://github.com/open-policy-agent/opa/issues/1508

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,11 @@ The OPA-Istio plugin supports the following configuration fields:
 | Field | Required | Description |
 | --- | --- | --- |
 | `plugins["envoy_ext_authz_grpc"].addr` | No | Set listening address of Envoy External Authorization gRPC server. This must match the value configured in the Envoy Filter resource. Default: `:9191`. |
-| `plugins["envoy_ext_authz_grpc"].query` | No | Specifies the name of the policy decision to query. The policy decision must be return a `boolean` value. `true` indicates the request should be allowed and `false` indicates the request should be denied. Default: `data.istio.authz.allow`. |
+| `plugins["envoy_ext_authz_grpc"].query` | No | Specifies the name of the policy decision to query. The policy decision can either be a `boolean` or an `object`. If boolean, `true` indicates the request should be allowed and `false` indicates the request should be denied. If the policy decision is an object, it **must** contain the `allowed` key set to either `true` or `false` to indicate if the request is allowed or not respectively. It can optionally contain a `headers` field to send custom headers to the downstream client or upstream. An optional `body` field can be included in the policy decision to send a response body data to the downstream client. Also an optional `http_status` field can be included to send a HTTP response status code to the downstream client other than `403 (Forbidden)`. Default: `data.istio.authz.allow`.|
+
+If the configuration does not specify the `query` field, `data.istio.authz.allow` will be considered as the default name of the policy decision to query.
+
+An example of a rule that returns an object that not only indicates if a request is allowed or not but also provides optional response headers, body and HTTP status that can be sent to the downstream client or upstream can be seen below in the [Example Policy with Object Response](#example-policy-with-object-response) section.
 
 In the [Quick Start](#quick-start) section an OPA policy is loaded via a volume-mounted ConfigMap. For production deployments, we recommend serving policy [Bundles](http://www.openpolicyagent.org/docs/bundles.html) from a remote HTTP server. For example:
 
@@ -277,6 +281,29 @@ default allow = false
 
 allow {
    input.parsed_path = ["api", "v1", "products"]
+}
+```
+
+## Example Policy with Object Response
+
+The `allow` rule in the below policy when queried generates an `object` that provides the status of the request (ie. `allowed` or `denied`) alongwith some headers, body data and HTTP status which will be included in the response that is sent back to the downstream client or upstream.
+
+```ruby
+package istio.authz
+
+default allow = {
+  "allowed": false,
+  "headers": {"x-ext-auth-allow": "no"},
+  "body": "Unauthorized Request",
+  "http_status": 301
+}
+
+allow = response {
+  input.attributes.request.http.method == "GET"
+  response := {
+    "allowed": true,
+    "headers": {"x-ext-auth-allow": "yes"}
+  }
 }
 ```
 
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -16,7 +16,9 @@ import (
 
 	ctx "golang.org/x/net/context"
 
+	ext_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	ext_authz "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
+	ext_type "github.com/envoyproxy/go-control-plane/envoy/type"
 	google_rpc "github.com/gogo/googleapis/google/rpc"
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/metrics"
@@ -26,6 +28,7 @@ import (
 	"github.com/open-policy-agent/opa/server"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/util"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 )
@@ -39,7 +42,7 @@ type evalResult struct {
 	revision   string
 	decisionID string
 	txnID      uint64
-	decision   bool
+	decision   interface{}
 	metrics    metrics.Metrics
 }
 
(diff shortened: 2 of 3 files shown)


internal/internal.go
```diff
@@ -263,6 +330,71 @@ func (p *envoyExtAuthzGrpcServer) log(ctx context.Context, input interface{}, re
...
+	}
+
+	if decision {
+		status = int32(google_rpc.OK)
+	}
+
+	return status, nil
+}
+
+func getResponeHeaders(result map[string]interface{}) ([]*ext_core.HeaderValueOption, error) {
```
patrick-east: nit: typo `Respone` -> `Response`

README.md
```diff
@@ -133,7 +133,13 @@ The OPA-Istio plugin supports the following configuration fields:
...
+| `plugins["envoy_ext_authz_grpc"].boolean_query` | No | Specifies the name of the policy decision to query. The policy decision must return a `boolean` value. `true` indicates the request should be allowed and `false` indicates the request should be denied. Default: `data.istio.authz.allow`.|
+| `plugins["envoy_ext_authz_grpc"].response_query` | No | Specifies the name of the policy decision to query. The policy decision must be an object containing the `status` key set to either `true` or `false` to indicate if the request is allowed or not respectively. The policy decision can optionally contain a `headers` field to send custom headers to the downstream client or upstream. Also an optional `body` field can be included in the policy decision to send a response body data to the downstream client.|
+
+The `boolean_query` (or `query`) and `response_query` fields are mutually exclusive. We recommend using the `boolean_query` field instead of the soon to be deprecated `query` field.
```
patrick-east: nit: Should we maybe say something like "We recommend using the `boolean_query` or `response_query` field instead of the soon to be deprecated `query` field"? If I understand correctly we just want people to switch to the newer config options, right?

internal/internal.go
ashutosh-narkar: 👍

README.md
ashutosh-narkar: `boolean_query` and `response_query` are the new fields. `boolean_query` is the replacement for the `query` option.  Will re-word this sentence.

tsandall:
I forget where we landed on this but I believe we discussed just keeping the `query` parameter and extending the gRPC server to accept non-boolean responses that contain the fields needed for the Check response. The benefit of this approach is that if users need to migrate from boolean to non-boolean, they don't have to reconfigure the OPA, they just update their policy and it'll work.

ashutosh-narkar:
Yes this is correct. What we also discussed is to have :

`boolean_query` which returns true/false
`response_query` which returns HTTP response body, status code, headers.

We discussed to deprecate the `query` field in favor of `boolean_query`.

tsandall:
Having two separate parameters (and deprecating the existing one) was the initial plan but I thought we landed on making query support both boolean and response formats (in which case we don't have to deprecate anything, it's purely additive.)

ashutosh-narkar:
The plan would be to keep the current `query` parameter in the config. The decision that the query specifies can return a result that is either `boolean` or `object`. If the decision is an `object`, it should contain the `status` key set to either `true` or `false` to indicate if the request is allowed or not respectively. The policy decision can optionally contain a `headers` field to send custom headers to the downstream client or upstream. Also an optional `body` field can be included in the policy decision to send a response body data to the downstream client.

We would then return the appropriate `CheckResponse ` based on the query result type.

Does that sound ok @tsandall ?

tsandall:
That sounds good to me. cc @patrick-east this changes the PR a bit but I think it's the right approach.

ashutosh-narkar:
@tsandall updated the PR to keep the current `query` field. The policy decision can be either a `boolean` or an `object`.

danielpops:
I think it's a really good decision to maintain the simple "query" interface and then have the logic in Opa to do "the right thing". Thanks @ashutosh-narkar and @tsandall for this! These changes look really good!

FYI I recently had a [PR](https://github.com/Yelp/opa-istio-plugin/pull/2) in a fork of this repo to add a `--dry-run` flag, which I think would still make sense in the context of these changes. Once this PR is merged, I'll look to re-base the dry-run changes on top of this and have you guys review it.

internal/internal.go
```diff
@@ -263,6 +302,71 @@ func (p *envoyExtAuthzGrpcServer) log(ctx context.Context, input interface{}, re
...
+}
+
+func getResponseBody(result map[string]interface{}) (string, error) {
+	var ok bool
+	var val interface{}
+
+	if val, ok = result["body"]; !ok {
+		return "", nil
+	}
+	return fmt.Sprintf("%v", val), nil
```
tsandall: I think we expect the policy decision to represent the body as a string. Sprintfing is going to break for things like maps/slices/etc.

internal/internal.go
```diff
@@ -263,6 +302,71 @@ func (p *envoyExtAuthzGrpcServer) log(ctx context.Context, input interface{}, re
...
+	}
+
+	if headers, ok = val.(map[string]interface{}); !ok {
+		return nil, fmt.Errorf("type assertion error")
+	}
+
+	for key, value := range headers {
+		headerValue := &ext_core.HeaderValue{
+			Key:   key,
+			Value: fmt.Sprintf("%v", value),
```
tsandall: I think we expect the values to be strings. Same comment about the body below.

internal/internal.go
```diff
@@ -153,14 +155,52 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
...
+		if decision {
+			status = int32(google_rpc.OK)
+		}
+
+		resp.Status = &google_rpc.Status{Code: status}
+
+	case map[string]interface{}:
+		status, err := getResponseStatus(decision)
+		if err != nil {
+			return nil, err
```
tsandall: I'd consider wrapping the error so that we can determine whether it was the status or the headers or the body that caused the error.

README.md
```diff
@@ -133,7 +133,11 @@ The OPA-Istio plugin supports the following configuration fields:
...
+| `plugins["envoy_ext_authz_grpc"].query` | No | Specifies the name of the policy decision to query. The policy decision can either be a `boolean` or an `object`. If boolean, `true` indicates the request should be allowed and `false` indicates the request should be denied. If the policy decision is an object, it **must** contain the `status` key set to either `true` or `false` to indicate if the request is allowed or not respectively. It can optionally contain a `headers` field to send custom headers to the downstream client or upstream. Also an optional `body` field can be included in the policy decision to send a response body data to the downstream client. Default: `data.istio.authz.allow`.|
+
+If the configuration does not specify the `query` field, `data.istio.authz.allow` will be considered as the default name of the policy decision to query. The policy decision returned by the rule specified by the default `query` is `boolean` in nature.
```
tsandall: I'd remove the last sentence as the semantics do not depend on the name.

internal/internal_test.go
```diff
@@ -251,73 +252,303 @@ func TestCheckWithLoggerError(t *testing.T) {
...
+			Query:       query,
+			parsedQuery: parsedQuery,
+		},
+		manager: m,
+	}
+	return s
+}
 
-	if err := m.Start(ctx); err != nil {
+func testAuthzServerwithObjectDecision(customLogger plugins.Plugin) *envoyExtAuthzGrpcServer {
```
danielpops: nit: camelCasing should be `testAuthzServerWithObjectDecision` (capital `With`)

internal/internal_test.go
```diff
@@ -251,73 +252,303 @@ func TestCheckWithLoggerError(t *testing.T) {
...
-	default allow = false
+	assertHeaders(t, headers, expectedHeaders)
+}
+
+func TestCheckDenyObjectDecision(t *testing.T) {
 
-	allow {
-		roles_for_user[r]
-		required_roles[r]
+	exampleRequest := `{
```
danielpops: Why does this test implement it's own input data, when there is the `exampleDeniedRequest` data used in all other `Deny` tests?

internal/internal.go
ashutosh-narkar: Removed `fmt.Sprintf`

internal/internal.go
ashutosh-narkar: Done. Removed `fmt.Sprintf`.

internal/internal.go
ashutosh-narkar: Wrapped errors to include more context.

README.md
ashutosh-narkar: Removed.

internal/internal_test.go
ashutosh-narkar: Done.

internal/internal_test.go
ashutosh-narkar: Good point. Updated the test to use the `exampleDeniedRequest` as input.

PR #103, opened 2019-07-30
Add a flag for dry-run mode, which configures Opa …

danielpops:
…to unconditionally return 
`ext_authz.CheckResponse.Status` => `google_rpc.Status{Code: google_rpc.OK}`
`ext_authz.CheckResponse.HttpReponse` => `ext_authz.OkHttpResponse`

This will allow for testing new policies without impacting live traffic. The intended use case is to run with `dry-run: true` and watch decision logs to see what "would" have happened.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,8 +134,11 @@ The OPA-Istio plugin supports the following configuration fields:
 | --- | --- | --- |
 | `plugins["envoy_ext_authz_grpc"].addr` | No | Set listening address of Envoy External Authorization gRPC server. This must match the value configured in the Envoy Filter resource. Default: `:9191`. |
 | `plugins["envoy_ext_authz_grpc"].query` | No | Specifies the name of the policy decision to query. The policy decision can either be a `boolean` or an `object`. If boolean, `true` indicates the request should be allowed and `false` indicates the request should be denied. If the policy decision is an object, it **must** contain the `allowed` key set to either `true` or `false` to indicate if the request is allowed or not respectively. It can optionally contain a `headers` field to send custom headers to the downstream client or upstream. An optional `body` field can be included in the policy decision to send a response body data to the downstream client. Also an optional `http_status` field can be included to send a HTTP response status code to the downstream client other than `403 (Forbidden)`. Default: `data.istio.authz.allow`.|
+| `plugins["envoy_ext_authz_grpc"].dry-run` | No | Configures the Envoy External Authorization gRPC server to unconditionally return an `ext_authz.CheckResponse.Status` of `google_rpc.Status{Code: google_rpc.OK}`. Default: `false`. |
 
 If the configuration does not specify the `query` field, `data.istio.authz.allow` will be considered as the default name of the policy decision to query.
+
+The `dry-run` parameter is provided to enable you to test out new policies. You can set `dry-run: true` which will unconditionally allow requests. Decision logs can be monitored to see what "would" have happened. This is especially useful for initial integration of OPA or when policies undergo large refactoring.
 
 An example of a rule that returns an object that not only indicates if a request is allowed or not but also provides optional response headers, body and HTTP status that can be sent to the downstream client or upstream can be seen below in the [Example Policy with Object Response](#example-policy-with-object-response) section.
 
@@ -156,8 +159,9 @@ bundle:
   service: bundle_service
 plugins:
     envoy_ext_authz_grpc:
-        addr: :9191
-        query: data.istio.authz.allow
+        addr:    :9191
+        query:   data.istio.authz.allow
+        dry-run: false
 ```
 
 ## Example Policy
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -35,6 +35,7 @@ import (
 
 const defaultAddr = ":9191"
 const defaultQuery = "data.istio.authz.allow"
+const defaultDryRun = false
 
 var revisionPath = storage.MustParsePath("/system/bundle/manifest/revision")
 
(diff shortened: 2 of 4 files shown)


internal/internal_test.go
```diff
@@ -380,6 +428,67 @@ func TestCheckDenyObjectDecision(t *testing.T) {
...
+	}
+
+	server := testAuthzServerWithObjectDecision(&testPlugin{}, true)
+	ctx := context.Background()
+	output, err := server.Check(ctx, &req)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	fmt.Printf("Result is %v\n", output)
```
ashutosh-narkar: Remove `fmt.Printf`

internal/internal_test.go
```diff
@@ -380,6 +428,67 @@ func TestCheckDenyObjectDecision(t *testing.T) {
...
+
+	if output.Status.Code != int32(google_rpc.OK) {
+		t.Fatalf("Expected request to be allowed since config.DryRun is true, but got: %v", output)
+	}
+
+	response := output.GetOkResponse()
+	if response == nil {
+		t.Fatal("Expected OkHttpResponse struct but got nil")
+	}
```
ashutosh-narkar: Related to comment about empty `HttpResponse `.

internal/internal.go
```diff
@@ -224,13 +229,25 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
...
 	}).Info("Returning policy decision.")
 
+	// If dry-run mode, override the Status code to unconditionally Allow the request
+	// DecisionLogging should reflect what "would" have happened
+	if p.cfg.DryRun {
+		if resp.Status.Code != int32(google_rpc.OK) {
+			resp.Status = &google_rpc.Status{Code: int32(google_rpc.OK)}
+			resp.HttpResponse = &ext_authz.CheckResponse_OkResponse{
+				OkResponse: &ext_authz.OkHttpResponse{},
+			}
```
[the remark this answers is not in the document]
danielpops: I kind of assumed that Envoy or istio or etc would use the `HttpResponse` field and that leaving it uninitialized (or set to DeniedHttpResponse) was not a good thing. My intention is to make sure that the request is authorized when `dry-run = true`. If the value of response.HttpResponse doesn't technically matter, I can remove this part of the code.

internal/internal_test.go
danielpops: This was copied from other tests :-p

Removed!

internal/internal.go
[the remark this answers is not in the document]
ashutosh-narkar: That's a good point. You don't want an unconditional allowed request have a response corresponding to a denied. This should be ok as is or you could set it to nil.

internal/internal_test.go
danielpops: See my other comment. lmk what you think.

internal/internal_test.go
ashutosh-narkar: Ohh no ! I will take care of `TestCheckDenyObjectDecision `

internal/internal.go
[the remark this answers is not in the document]
danielpops: ok. I'll leave it as-is for now.

README.md
```diff
@@ -134,6 +134,7 @@ The OPA-Istio plugin supports the following configuration fields:
...
 | `plugins["envoy_ext_authz_grpc"].query` | No | Specifies the name of the policy decision to query. The policy decision can either be a `boolean` or an `object`. If boolean, `true` indicates the request should be allowed and `false` indicates the request should be denied. If the policy decision is an object, it **must** contain the `allowed` key set to either `true` or `false` to indicate if the request is allowed or not respectively. It can optionally contain a `headers` field to send custom headers to the downstream client or upstream. An optional `body` field can be included in the policy decision to send a response body data to the downstream client. Also an optional `http_status` field can be included to send a HTTP response status code to the downstream client other than `403 (Forbidden)`. Default: `data.istio.authz.allow`.|
+| `plugins["envoy_ext_authz_grpc"].dry-run` | No | Configures the Envoy External Authorization gRPC server to unconditionally return an `ext_authz.CheckResponse.Status` of `google_rpc.Status{Code: google_rpc.OK}`. Default: `false`. |
```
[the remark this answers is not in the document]
danielpops: I added a note. Let me know if that sounds ok. Happy to reword or change entirely as you see fit :)

internal/internal.go
```diff
@@ -224,13 +229,25 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
...
 		"decision":            result.decision,
 		"err":                 err,
 		"txn":                 result.txnID,
 		"metrics":             result.metrics.All(),
 		"total_decision_time": time.Since(start),
 	}).Info("Returning policy decision.")
 
+	// If dry-run mode, override the Status code to unconditionally Allow the request
+	// DecisionLogging should reflect what "would" have happened
+	if p.cfg.DryRun {
```
tsandall: I think we should defer this block at the top of the function so that errors due not trigger denial when dry-run is enabled. /cc @ashutosh-narkar

ashutosh-narkar:
> In addition to the comment below we should figure out how we could indicate in the decision log that dry-run was enabled. It's not immediately clear how we can do this with current decision log message format.

One option would be to group all client related info such as `RemoteAddr`, `dry-run` etc. into a new struct and then update the `RequestedBy`(string) field in the decision log event to something like `ClientInfo`(object) to include client address, dry-run etc.

PR #115, opened 2019-08-03
Migrate to gomod

ashutosh-narkar:
This pull request contains changes for switching from `dep` to `modules`.


tsandall:
No major feedback from me. This is a case where some context that explains _why_ the changes are being made would be nice to have in the commit message.

ashutosh-narkar:
@tsandall update the commit message explaining why the migration is done.

PR #120, opened 2019-08-08
Use prepared queries

ashutosh-narkar:
These changes use prepared queries that cache compiled queries thereby reducing query compile latency.

Fixes: open-policy-agent/opa/issues/1553

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"net"
 	"strings"
+	"sync"
 	"time"
 
 	ctx "golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


internal/internal.go
```diff
@@ -279,16 +290,34 @@ func (p *envoyExtAuthzGrpcServer) eval(ctx context.Context, input ast.Value, opt
...
 
-		opts = append(opts,
-			rego.Metrics(result.metrics),
-			rego.ParsedQuery(p.cfg.parsedQuery),
-			rego.ParsedInput(input),
-			rego.Compiler(p.manager.GetCompiler()),
-			rego.Store(p.manager.Store),
-			rego.Transaction(txn),
-			rego.Runtime(p.manager.Info))
+		p.preparedQueryMutex.Lock()
```
tsandall: @ashutosh-narkar @patrick-east are we serializing policy evaluation intentionally here? I would have expected the lock to only be held when checking if the prepared query needs to be constructed. Assuming the Envoy and the ext_authz filter can execute concurrent checks, this will become a bottleneck. We could avoid the bottleneck by moving the locking/construction into a helper function.

PR #123, opened 2019-08-10
Set denied status when it exists.

ashutosh-narkar:
The denied status is a required field that should be set when a valid http satus is provided. Setting this to nil can cause Envoy to return an Unknown error on a denied request.


tsandall:
It would be good to cover this kind of change with a unit test.

ashutosh-narkar:
There is a unit test which covers `getResponseHTTPStatus` functionality. After system testing, it was discovered this change was not resolving the issue which caused Envoy to return an `Unknown` status code. The issue was resolved with PR https://github.com/open-policy-agent/opa-istio-plugin/pull/124. The change was both unit and system tested.

PR #125, opened 2019-08-13
Provide parsed body

ashutosh-narkar:
The pull request addresses the following: 

* Provide parsed body as input to OPA policy similar to parsed path.

* Avoid serializing policy evaluation.

See commit for more details.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -161,6 +161,7 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
 	}
 
 	input["parsed_path"] = getParsedPath(req)
+	input["parsed_body"] = getParsedBody(req)
 
 	inputValue, err := ast.InterfaceToValue(input)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/internal.go
```diff
@@ -505,3 +509,10 @@ func getParsedPath(req *ext_authz.CheckRequest) []interface{} {
 	}
 	return parsedPathInterface
 }
+
+func getParsedBody(req *ext_authz.CheckRequest) map[string]interface{} {
+	body := req.GetAttributes().GetRequest().GetHttp().GetBody()
+	var data map[string]interface{}
+	json.Unmarshal([]byte(body), &data)
```
tsandall: Need to call `UseNumber` on the JSON decoder here otherwise there is potential for loss of precision on large integers. If you want, you can use the util package in OPA which does it for you.

internal/internal_test.go
```diff
@@ -117,6 +145,57 @@ func TestCheckAllow(t *testing.T) {
 	}
 }
 
+func TestCheckTrigger(t *testing.T) {
+
+	// Example Mixer Check Request for input:
```
tsandall: Nit: s/Mixer/Envoy/

internal/internal.go
ashutosh-narkar: Used `util.Unmarshal`

internal/internal_test.go
ashutosh-narkar: Done

ashutosh-narkar:
@tsandall  Regarding cleaning up the tests created https://github.com/open-policy-agent/opa/issues/1627 to track it.

tsandall:
I just realized we're not checking the content-type in the request headers before unmarshaling the body. I'm guessing this is why we're ignoring the returned error? We should fix this to only unmarshal the JSON if the content-type indicates the message body is JSON and return an error in that case if unmarshaling fails.

PR #127, opened 2019-08-14
Check for errors while unmarshaling request body

ashutosh-narkar:
Previously unmarshaling errors were ignored and no check was performed to determine the request content-type. Errors are now returned if a json request body cannot be unmarshalled.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -160,7 +160,13 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
 	}
 
 	input["parsed_path"] = getParsedPath(req)
-	input["parsed_body"] = getParsedBody(req)
+
+	parsedBody, err := getParsedBody(req)
+	if err != nil {
+		return nil, err
+	}
+
+	input["parsed_body"] = parsedBody
 
 	inputValue, err := ast.InterfaceToValue(input)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/internal.go
```diff
@@ -510,9 +516,20 @@ func getParsedPath(req *ext_authz.CheckRequest) []interface{} {
...
+func getParsedBody(req *ext_authz.CheckRequest) (map[string]interface{}, error) {
 	body := req.GetAttributes().GetRequest().GetHttp().GetBody()
+	headers := req.GetAttributes().GetRequest().GetHttp().GetHeaders()
+
 	var data map[string]interface{}
-	util.Unmarshal([]byte(body), &data)
-	return data
+
+	if val, ok := headers["content-type"]; ok {
+		if val == "application/json" {
```
tsandall: We should probably check if the content-type _contains_ application/json instead of strict equality because if there are parameters then this will fail.

internal/internal_test.go
```diff
@@ -757,33 +760,84 @@ func TestGetResponseHttpStatus(t *testing.T) {
...
+			  },
+			  "body": "{\"firstname\": \"foo\", \"lastname\": \"bar\"}"
+			}
+		  }
+		}
+	  }`
+
+	tests := map[string]struct {
+		input *ext_authz.CheckRequest
+		want  int
```
tsandall: Perhaps we should just assert on the body content here? It's not clear why we would want to assert on the number of elements.

internal/internal_test.go
```diff
@@ -913,6 +967,14 @@ func testAuthzServerWithObjectDecision(customLogger plugins.Plugin, dryRun bool)
 	return s
 }
 
+func createCheckRequest(module string) *ext_authz.CheckRequest {
```
tsandall: Nit: I'd rename this parameter to reflect it's purpose (looks like it was copy-pasted from below?)

internal/internal.go
ashutosh-narkar: Done.

internal/internal_test.go
ashutosh-narkar: Updated test to assert body.

internal/internal_test.go
ashutosh-narkar: Done

PR #129, opened 2019-08-15
Add prepocessed path query to input. Fixes #1636

vetyy:
Signed-off-by: Matej Vetrak <matej.vetrak@exponea.com>

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/internal.go
```diff
@@ -506,8 +508,32 @@ func getRevision(ctx context.Context, store storage.Store, txn storage.Transacti
 	return revision, nil
 }
 
-func getParsedPath(req *ext_authz.CheckRequest) []interface{} {
+func getParsedQuery(req *ext_authz.CheckRequest) map[string]interface{} {
 	path := req.GetAttributes().GetRequest().GetHttp().GetPath()
+
+	parsedQuery := make(map[string]interface{})
```
tsandall: Let's use `net/url` to parse the URL here instead of rolling it from scratch. A side of effect of updating to use `net/url` is that the URL parameter values will be arrays of strings instead of just strings. This can't be avoided though (e.g., ?x=1&x=2 is valid.) We should also be able to parse the string once and construct the parsed path and query params at the same time.

internal/internal_test.go
```diff
@@ -94,7 +94,7 @@ const exampleAllowedRequestParsedPath = `{
 		"http": {
 		  "id": "13359530607844510314",
 		  "method": "GET",
-		  "path": "/my/test/path"
+		  "path": "/my/test/path?a=b&x=y"
```
tsandall: Tests exercising URL encoding should be added.

vetyy:
Thanks for the review. I've tried to fix the issues. Let me know if something needs to be changed/added.

PR #133, opened 2019-08-16
Downgrade request logging

ashutosh-narkar:
These changes change request logging from Info to Debug. Earlier the full request was looged at Info which is unnecessary especially when the console decision logger can be used instead.

Fixes #open-policy-agent/opa/issues/1554

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -138,7 +138,7 @@ func (p *envoyExtAuthzGrpcServer) listen() {
 	// The listener is closed automatically by Serve when it returns.
 	l, err := net.Listen("tcp", p.cfg.Addr)
 	if err != nil {
-		logrus.WithField("err", err).Fatalf("Unable to create listener.")
+		logrus.WithField("err", err).Fatal("Unable to create listener.")
 	}
 
 	logrus.WithFields(logrus.Fields{
@@ -146,10 +146,10 @@ func (p *envoyExtAuthzGrpcServer) listen() {
 		"query":             p.cfg.Query,
 		"dry-run":           p.cfg.DryRun,
 		"enable-reflection": p.cfg.EnableReflection,
-	}).Infof("Starting gRPC server.")
+	}).Info("Starting gRPC server.")
 
 	if err := p.server.Serve(l); err != nil {
-		logrus.WithField("err", err).Fatalf("Listener failed.")
+		logrus.WithField("err", err).Fatal("Listener failed.")
 	}
 
 	logrus.Info("Listener exited.")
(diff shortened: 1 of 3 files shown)


internal/internal.go
```diff
@@ -138,21 +138,21 @@ func (p *envoyExtAuthzGrpcServer) listen() {
...
+		logrus.WithField("err", err).Fatal("Unable to create listener.")
 	}
 
 	logrus.WithFields(logrus.Fields{
 		"addr":              p.cfg.Addr,
 		"query":             p.cfg.Query,
 		"dry-run":           p.cfg.DryRun,
 		"enable-reflection": p.cfg.EnableReflection,
-	}).Infof("Starting gRPC server.")
+	}).Debug("Starting gRPC server.")
```
tsandall: These logs don't have to be debug level (and they shouldn't be since they're important for knowing whether the listener started, with what configuration, etc.) Same w/ the exit log below.

internal/internal.go
```diff
@@ -265,7 +265,7 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckReq
 		"txn":                 result.txnID,
 		"metrics":             result.metrics.All(),
 		"total_decision_time": time.Since(start),
-	}).Info("Returning policy decision.")
+	}).Debug("Returning policy decision.")
```
tsandall: Did you check the improvement from this? I could imagine making this and the debug log below conditional if the field construction is costly.

```
if logrus.debugenabled {
    // emit log
}
```

internal/internal.go
ashutosh-narkar: Changed log level to `Info`.

internal/internal.go
ashutosh-narkar: @tsandall Here are the benchmark results:

**With Debug log** 
```bash
$ go test -run=XXX -bench=.
goos: darwin
goarch: amd64
pkg: github.com/open-policy-agent/opa-istio-plugin/internal
BenchmarkCheck-8       10000            248507 ns/op
PASS
ok      github.com/open-policy-agent/opa-istio-plugin/internal  2.576s
```

**Without Debug log** 
```bash
$ go test -run=XXX -bench=.
goos: darwin
goarch: amd64
pkg: github.com/open-policy-agent/opa-istio-plugin/internal
BenchmarkCheck-8     10000            222931 ns/op
PASS
ok      github.com/open-policy-agent/opa-istio-plugin/internal  2.317s
```

So the improvement is around 25 microseconds.

ashutosh-narkar: Added a check to see if debug logging is enabled.

PR #137, opened 2019-09-16
Add system test to validate end-to-end functionality.

ashutosh-narkar:
These changes add a end-to-end test that sanity checks the integration between istio-proxy and opa-istio-plugin.

Fixes #open-policy-agent/opa/issues/1510

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: required
 language: go
+go_import_path: github.com/open-policy-agent/opa-istio-plugin
 go:
 - '1.11'
 services:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -9,10 +10,38 @@ env:
...
+      if: (NOT type IN (pull_request)) AND (branch = master)
+      before_script:
+      - KIND_VERSION=0.4.0
+      - ISTIO_VERSION=1.2.5
+      # Download and install kind
+      - curl -L https://github.com/kubernetes-sigs/kind/releases/download/v${KIND_VERSION}/kind-linux-amd64 --output kind && chmod +x kind && sudo mv kind /usr/local/bin/
+      # Create kind cluster
+      - |
+        if [ $(kind get clusters) ]; then
+          export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
```
patrick-east: If I'm interpreting this correctly, this lets it re-use clusters? Do we want to allow that for testing? I ask because I didn't see anything in the tests or script to cleanup afterwards.

.travis.yml
```diff
@@ -9,10 +10,38 @@ env:
...
+        - provider: script
+          skip_cleanup: true
+          script: make deploy-travis
+          on:
+            repo: open-policy-agent/opa-istio-plugin
+            branch: master
+            tags: false
+    - stage: e2e
+      if: (NOT type IN (pull_request)) AND (branch = master)
+      before_script:
```
patrick-east: Is it possible to put this stuff in `before_script` as a make target and/or shell script in the repo? Seems like it would be pretty useful to have a like one-liner that sets up a kind cluster with istio (with the correct versions) for local dev/test of the plugin.

test/bats/test.bats
```diff
@@ -0,0 +1,82 @@
...
+}
+
+@test "alice can access /productpage" {
+  host="$(docker_ip 'kind-control-plane')"
+  port="$(ingress_port)"
+
+  echo "$host"
+  echo "$port"
+
+  result="$(docker exec kind-control-plane curl --connect-timeout 30 --max-time 60 --retry 10 --retry-delay 0 --retry-max-time 600 --retry-connrefused --user alice:password -s -o /dev/null -w "%{http_code}" http://"$host":"$port"/productpage)"
```
patrick-east: It might make sense to have a variable in the helpers (or a function) that can do the curl's with all the like timeouts, retries, etc defined. Would simplify these lines quite a bit to make it easier to see what is being tested for.

At a minimum I'd request that the line get broken up with `\` so there isn't any horizontal scrolling needed on narrow views 😄

.travis.yml
```diff
@@ -9,10 +10,38 @@ env:
...
+      deploy:
+        - provider: script
+          skip_cleanup: true
+          script: make deploy-travis
+          on:
+            repo: open-policy-agent/opa-istio-plugin
+            branch: master
+            tags: false
+    - stage: e2e
+      if: (NOT type IN (pull_request)) AND (branch = master)
```
patrick-east: More of a question about travis: when would this be triggered to run? Post merge to master? Or do we have a CRON trigger somewhere?

.travis.yml
ashutosh-narkar: Added:
```yaml
after_script:
      - kind delete cluster
```

.travis.yml
ashutosh-narkar: Created a make target `test-cluster` which creates a Kubernetes cluster using kind and deploys Istio on top of it. But when I included this target as part of Travis's `before_script`, the Travis `script` (ie. `make test-e2e`) fails and is unable to connect to the Kube cluster. 

Same result is seen if `make test-cluster` is made part of the Travis `script`. So I'm going to keep the `before_script` as is for now and we can re-visit this later.  Also keeping the new make target `test-cluster` for local testing.

test/bats/test.bats
ashutosh-narkar: Broke up the lines to assist visibility.

.travis.yml
ashutosh-narkar: This would run post merge to master.

PR #139, opened 2019-09-19
Remove support for Go plugin.

ashutosh-narkar:
With OPA v0.14.0 deprecating support for Go plugins, opa-istio will no longer be built as a Go plugin. These changes customize OPA by leveraging the custom plugin framework.


cmd/opa-istio-plugin/main.go
```diff
@@ -1,11 +1,15 @@
-// Copyright 2018 The OPA Authors. All rights reserved.
+// Copyright 2018 The OPA Authors.  All rights reserved.
```
patrick-east: nit: extra space added in?

ashutosh-narkar: Fixed.

PR #144, opened 2019-10-07
Add required stat_prefix on Envoy authz network filter

teh-username:
Hello,

This is to fix the following error I get whenever I try to apply the `quick_start.yaml` as is:

```
Proto constraint validation failed (ExtAuthzValidationError.StatPrefix: ["value length must be at least " '\x01' " bytes"]): grpc_service {
  google_grpc {
    target_uri: "127.0.0.1:9191"
    stat_prefix: "ext_authz"
  }
}
, virtualInbound: Proto constraint validation failed (ExtAuthzValidationError.StatPrefix: ["value length must be at least " '\x01' " bytes"]): grpc_service {
  google_grpc {
    target_uri: "127.0.0.1:9191"
    stat_prefix: "ext_authz"
  }
}
```

As per the [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/ext_authz/v2/ext_authz.proto#network-external-authorization), `stat_prefix` is required on the same level as `grpc_service` in the network filter part of the `EnvoyFilter` object. 

In case anyone is curious, I am using Istio 1.3.1.

### DIFF
--- a/quick_start.yaml
+++ b/quick_start.yaml
@@ -36,6 +36,7 @@ spec:
     filterType: NETWORK
     filterName: "envoy.ext_authz"
     filterConfig:
+      stat_prefix: "ext_authz"
       grpc_service:
         google_grpc:
           target_uri: 127.0.0.1:9191


ashutosh-narkar:
@teh-username in `quick_start.yaml`, the Envoy filter type is `HTTP`. `stat_prefix` is not a required field as per the [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/http/ext_authz/v2/ext_authz.proto.html?highlight=with_request_body).

Did you change the filter type ?

teh-username:
Hi @ashutosh-narkar,

The `quick_start.yaml` in the `master` branch injects both `HTTP` and `NETWORK` filter type. My setup involves a python application and a redis service which is maybe why it doesn't work? Since it's the `NETWORK` filter that's catching the traffic.

I used this [quick link](https://github.com/open-policy-agent/opa-istio-plugin/blob/master/quick_start.yaml) from the README which has the following:

```
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: ext-authz
  namespace: istio-system
spec:
  filters:
  - insertPosition:
      index: FIRST
    listenerMatch:
      listenerType: SIDECAR_INBOUND
      listenerProtocol: HTTP
    filterType: HTTP
    filterName: "envoy.ext_authz"
    filterConfig:
      with_request_body:
        max_request_bytes: 8192
        allow_partial_message: true
      grpc_service:
        # NOTE(tsandall): when this was tested with the envoy_grpc client the gRPC
        # server was receiving check requests over HTTP 1.1. The gRPC server in
        # OPA-Istio would immediately close the connection and log that a bogus
        # preamble was sent by the client (it expected HTTP 2). Switching to the
        # google_grpc client resolved this issue.
        google_grpc:
          target_uri: 127.0.0.1:9191
          stat_prefix: "ext_authz"
  - insertPosition:
      index: FIRST
    listenerMatch:
      listenerType: SIDECAR_INBOUND
      listenerProtocol: TCP
    filterType: NETWORK
    filterName: "envoy.ext_authz"
    filterConfig:
      grpc_service:
        google_grpc:
          target_uri: 127.0.0.1:9191
          stat_prefix: "ext_authz"
```

Thanks

PR #150, opened 2019-12-02
update the tag#

mjschmidt:
matching up to docker hub. There is no 0.15.0-istio container right now


mjschmidt:
This was solved in the latest 0.15.1 pr

PR #151, opened 2019-12-03
updating the apiVersion of Deployment

LokeshAggarwal1997:
The apiVersion:  extension/v1beta1 is deprecated and eventually removed.


ashutosh-narkar:
@LokeshAggarwal1997 can you fix the dco.

LokeshAggarwal1997:
yes , looking into it

PR #159, opened 2020-01-24
config: Add a new parameter that specifies the policy decision path

ashutosh-narkar:
These changes deprecate the "query" parameter and add a new "path" parameter to specify the policy decision path. Earlier the query name was used to generate the policy decision path in decision logs. With changes to the decision logger api to no longer modify the decision path, the plugin now sets the appropriate value for the path and query in the decison log.

Fixes : open-policy-agent/opa/issues/2027

### DIFF
--- a/internal/internal_test.go
+++ b/internal/internal_test.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"testing"
 
@@ -275,7 +276,7 @@ func TestCheckAllowWithLogger(t *testing.T) {
 
 	event := customLogger.events[0]
 
-	if event.Error != nil || event.Query != "data.istio.authz.allow" || event.Path != "istio/authz/allow" || event.Revision != "" || *event.Result == false {
+	if event.Error != nil || event.Path != "istio/authz/allow" || event.Revision != "" || *event.Result == false {
 		t.Fatal("Unexpected events:", customLogger.events)
 	}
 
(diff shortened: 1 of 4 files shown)


internal/internal_test.go
```diff
@@ -545,6 +546,41 @@ func TestConfigValid(t *testing.T) {
...
+	tests := map[string]struct {
+		path string
+		want string
+	}{
+		"empty_path":               {path: "", want: "data.istio.authz.allow"},
+		"path_no_lt_slash":         {path: "test/allow/main", want: "data.test.allow.main"},
+		"path_with_leading_slash":  {path: "/test/allow/main", want: "data.test.allow.main"},
+		"path_with_trailing_slash": {path: "test/allow/main/", want: "data.test.allow.main"},
+		"path_with_lt_slash":       {path: "/test/allow/main/", want: "data.test.allow.main"},
+		"path_with_periods":        {path: "test/com.foo.istio.ingress/allow/main/", want: "data.test.com.foo.istio.ingress.allow.main"},
```
patrick-east: I'm not sure that this is correct, shouldn't it be something more like `data.test["com.foo.istio.ingress"].main` (well, a ref with like `["data", "test", "com.foo.istio.ingress", "main]`, we don't need that intermediate string representation)

I think we need to do something more like what the server does over here: https://github.com/open-policy-agent/opa/blob/master/server/server.go#L2179-L2205 (maybe don't need to do the url escape stuff for this one) rather than the swap of all `/` to `.` and parsing the resulting string. It should split on `/` and build the ref from the segments.

ashutosh-narkar:
@patrick-east Updated the path generation logic to resemble the server.

PR #175, opened 2020-04-24
fix parsing URL path

vetyy:
Hello,

I found a problem with parsing path and query parameters.

The original test is actually invalid and is escaping reserved characters '=', '?' and '&' as per (https://tools.ietf.org/html/rfc3986#section-2.2).

If I change test to include a '\n' character in query parameter it will be first unescaped by `url.PathUnescape` and then again using `url.Parse` that will result in error "net/url: invalid control character in URL".

So there is no need to explicitly unescape path as `url.Parse` handles that internally.

I updated the code and test to be valid and I also tested in our own OPA deployment.

Thanks,
Matej

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -565,12 +565,7 @@ func getRevision(ctx context.Context, store storage.Store, txn storage.Transacti
 func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[string]interface{}, error) {
 	path := req.GetAttributes().GetRequest().GetHttp().GetPath()
 
-	unescapedPath, err := url.PathUnescape(path)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	parsedURL, err := url.Parse(unescapedPath)
+	parsedURL, err := url.Parse(path)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/internal/internal_test.go
+++ b/internal/internal_test.go
@@ -1286,9 +1286,9 @@ func TestParsedPathAndQuery(t *testing.T) {
 			map[string]interface{}{"a": []interface{}{"1"}, "b": []interface{}{"2"}},
 		},
 		{
-			createExtReqWithPath("%2Fmy%2Ftest%2Fpath%3Fa%3D1%26a%3D2"),
+			createExtReqWithPath("/my/test/path?a=1&a=new%0aline"),
 			[]interface{}{"my", "test", "path"},
-			map[string]interface{}{"a": []interface{}{"1", "2"}},
+			map[string]interface{}{"a": []interface{}{"1", "new\nline"}},
 		},
 	}
 


ashutosh-narkar:
Hello @vetyy , can you help me understand this statement:
> The original test is actually invalid and is escaping reserved characters '=', '?' and '&' as per (https://tools.ietf.org/html/rfc3986#section-2.2).

Afaik both reserved and unreserved characters are allowed in an URI. While using unreserved characters, they are url-encoded, hence the code uses `url.PathUnescape`. The implementation of `url.Parse` and  `url.PathUnescape` are not quite the same.

So both these seem valid to me:

* `/my/test/path?a=1&a=new%0aline` => 

_decoded_
```
/my/test/path?a=1&a=new
line
```

* `%2Fmy%2Ftest%2Fpath%3Fa%3D1%26a%3D2` =>


_decoded_
```
/my/test/path?a=1&a=2
```

We should try to support both these if feasible imo.

vetyy:
I guess I was in a hurry and shouldn't have to delete the original test, but the expected behaviour from any webserver I tried is that encoded characters are treated as either part of the path or query parameters based on usage of the reserved characters, to demostrate what I mean and how webservers will actually handle the requests I'll try to show some tests and if it looks ok I can add them to the code:

```
// All characters are encoded
{
    createExtReqWithPath("%2Fmy%2Ftest%2Fpath%3Fa%3D1%26a%3D2"),
    []interface{}{"my", "test", "path?a=1&a=2"},
    map[string]interface{}{},
},
// '?" characters is not encoded
{
    createExtReqWithPath("%2Fmy%2Ftest%2Fpath?a%3D1%26a%3D2"),
    []interface{}{"my", "test", "path"},
    map[string]interface{}{"a=1&a=2": []interface{}{""}},
},
// '?' and first '=' is not encoded
{
    createExtReqWithPath("%2Fmy%2Ftest%2Fpath?a=1%26a%3D2"),
    []interface{}{"my", "test", "path"},
    map[string]interface{}{"a": []interface{}{"1&a=2"}},
},
// '?' , '=' and '&' is not encoded
{
    createExtReqWithPath("%2Fmy%2Ftest%2Fpath?a=1&a%3D2"),
    []interface{}{"my", "test", "path"},
    map[string]interface{}{"a": []interface{}{"1"}, "a=2": []interface{}{""}},
},
```

ashutosh-narkar:
@vetyy any update on this ?  Let me know if you need any help.

homeles:
I'm facing this issue as well, when the plugin is parsing Prometheus queries with `\n` on it. 

Is this going to be considered anytime soon?

ashutosh-narkar:
@homeles feel free to take this on and submit changes if you would like. Otherwise we could take this on over the next couple of weeks or so.

tsandall:
Closing due to inactivity. We can revisit this in the future as needed.

srenatus:
This has come up again (see link ☝️). I think the fix here is the correct one, and we should try to get this in.

@vetyy Would you mind having another look? Or would you be OK if I took over and kept your commit intact?

srenatus:
@vetyy I've taken your commit in got it merged in #227. I hope you're OK with this. Thanks again for bringing this up and providing a fix 🎉 😃

PR #177, opened 2020-05-02
Update the package path used in the LDFLAGS to set the version inform…

danielpops:
…ation correctly with go 1.13. Fixes open-policy-agent/opa#2374

See open-policy-agent/opa#2374 for more details and a demonstration of this fix working

cc @ashutosh-narkar FYI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,12 +20,10 @@ BUILD_COMMIT := $(shell ./build/get-build-commit.sh)
 BUILD_TIMESTAMP := $(shell ./build/get-build-timestamp.sh)
 BUILD_HOSTNAME := $(shell ./build/get-build-hostname.sh)
 
-OPA_VENDOR := vendor/github.com/open-policy-agent/opa
-
-LDFLAGS := "-X github.com/open-policy-agent/opa-istio-plugin/$(OPA_VENDOR)/version.Version=$(VERSION) \
-	-X github.com/open-policy-agent/opa-istio-plugin/$(OPA_VENDOR)/version.Vcs=$(BUILD_COMMIT) \
-	-X github.com/open-policy-agent/opa-istio-plugin/$(OPA_VENDOR)/version.Timestamp=$(BUILD_TIMESTAMP) \
-	-X github.com/open-policy-agent/opa-istio-plugin/$(OPA_VENDOR)/version.Hostname=$(BUILD_HOSTNAME)"
+LDFLAGS := "-X github.com/open-policy-agent/opa/version.Version=$(VERSION) \
+	-X github.com/open-policy-agent/opa/version.Vcs=$(BUILD_COMMIT) \
+	-X github.com/open-policy-agent/opa/version.Timestamp=$(BUILD_TIMESTAMP) \
+	-X github.com/open-policy-agent/opa/version.Hostname=$(BUILD_HOSTNAME)"
 
 GO15VENDOREXPERIMENT := 1
 export GO15VENDOREXPERIMENT


ashutosh-narkar:
Thanks @danielpops for submitting this change !

Have you tried this change with older versions of Go (< 1.13) ? Is the version information populated ?

If not, would we need to force module support and use of the vendor directory ie. `GO := GO111MODULE=on GOFLAGS=-mod=vendor go` to make it work with v1.12, v1.13 etc.

danielpops:
@ashutosh-narkar the master branch does not even build with go < 1.13 =>

```
root@258471848bab:/# go get github.com/open-policy-agent/opa-istio-plugin
package github.com/open-policy-agent/opa-istio-plugin: no Go files in /root/go/src/github.com/open-policy-agent/opa-istio-plugin
root@258471848bab:/# cd /root/go/src/github.com/open-policy-agent/opa-istio-plugin/
root@258471848bab:~/go/src/github.com/open-policy-agent/opa-istio-plugin# make build
go generate ./...
go build -o opa_istio_linux_amd64 -ldflags "-X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Version=0.19.2-istio -X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Vcs=5d7ba4b8 -X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Timestamp=2020-05-04T15:47:07Z -X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Hostname=258471848bab" ./cmd/opa-istio-plugin/...
# github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/tester
vendor/github.com/open-policy-agent/opa/tester/runner.go:483:5: b.ReportMetric undefined (type *testing.B has no field or method ReportMetric)
Makefile:55: recipe for target 'build' failed
make: *** [build] Error 2
root@258471848bab:~/go/src/github.com/open-policy-agent/opa-istio-plugin# go version
go version go1.12.7 linux/amd64
```
I also presumed that go >= 1.13 was required as of https://github.com/open-policy-agent/opa/commit/e6727cbdf2dfb82d8921f7202d800af08c7d265a

So I'm not sure if backwards compatibility is necessary here, given that master only builds with go >= 1.13

danielpops:
> If not, would we need to force module support and use of the vendor directory ie. `GO := GO111MODULE=on GOFLAGS=-mod=vendor go` to make it work with v1.12, v1.13 etc.

I may be misinterpreting your point, and/or there may be more than just this simple change, but that doesn't change the build results => 

```
root@91b83089cce1:~/go/src/github.com/open-policy-agent/opa-istio-plugin# git diff
diff --git a/Makefile b/Makefile
index d99d04a0..51d6d54b 100644
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ VERSION := $(shell ./build/get-opa-version.sh)-istio$(shell ./build/get-plugin-r

 PACKAGES := $(shell go list ./.../ | grep -v 'vendor')

-GO := go
+GO := GO111MODULE=on GOFLAGS=-mod=vendor go
 GOARCH := $(shell go env GOARCH)
 GOOS := $(shell go env GOOS)
 DISABLE_CGO := CGO_ENABLED=0
root@91b83089cce1:~/go/src/github.com/open-policy-agent/opa-istio-plugin# make build
GO111MODULE=on GOFLAGS=-mod=vendor go generate ./...
GO111MODULE=on GOFLAGS=-mod=vendor go build -o opa_istio_linux_amd64 -ldflags "-X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Version=0.19.2-istio -X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Vcs=5d7ba4b8-dirty -X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Timestamp=2020-05-04T15:53:42Z -X github.com/open-policy-agent/opa-istio-plugin/vendor/github.com/open-policy-agent/opa/version.Hostname=91b83089cce1" ./cmd/opa-istio-plugin/...
# github.com/open-policy-agent/opa/tester
vendor/github.com/open-policy-agent/opa/tester/runner.go:483:5: b.ReportMetric undefined (type *testing.B has no field or method ReportMetric)
Makefile:55: recipe for target 'build' failed
make: *** [build] Error 2
```

patrick-east:
If you include:
```diff
-GO := go
+GO := GO111MODULE=on GOFLAGS=-mod=vendor go
```
and leave the original paths with `github.com/open-policy-agent/opa-istio-plugin/$(OPA_VENDOR)/version.Version`  it might fix things.

I agree though that supporting <1.13 is probably a no-go with the way we integrate OPA as we have a dependency on the couple of packages in OPA that require it (as seen by the error messages @danielpops posted)

Arguably we should be using those go flags anyway to force that the `./vendor` directory is being used (not that we don't trust go modules, but having the full source included was a decision we made for reproducibility which kinda goes out the window if we don't actually use those source files).

danielpops:
@patrick-east see my prior comment. In the master branch `make build` fails with go < 1.13, even with the proposed flag changes, because of the backwards incompatibility in the test code re: `b.ReportMetric`.

I agree that something more has to change here, but I think that it is currently doing the wrong thing when building with go >= 1.13, and that shipping my PR as is is a good first step. It sounds like the decision around supporting multiple versions and the vendor approach vs having the full source code is a decision that will likely take more discussion and iteration, and I feel that this can be done as a follow-up to fixing the broken version identifier issue. 

Thoughts?

patrick-east:
Right.. I agreed with you that supporting <1.13 doesn't make sense.

I tested it out and even with forcing to use -mod=vendor we need to specify the original OPA path, so conclusion to

> it might fix things.

Is that no, it doesn't.

So it seems like either way we need to adjust those paths.

danielpops:
sgtm

PR #178, opened 2020-05-13
Update health probes

patrick-east:
This combines a few changes related to health probes for the plugins.

* Updates to using the new `--diagnostic-addr` OPA option to serve /health on a separate port
* Adds/modifies readiness and liveness probes to use the alternate port in the quickstart and readme
* Updates the health check URL to use `?plugins` and updates the gRPC plugin to report its status to the OPA plugin manager.

There are a couple random fixes included that I found along the way while working on the changes and testing them all out.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -22,6 +22,13 @@ import (
 	ext_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	ext_authz "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
 	ext_type "github.com/envoyproxy/go-control-plane/envoy/type"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/genproto/googleapis/rpc/code"
+	rpc_status "google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+
 	"github.com/open-policy-agent/opa/ast"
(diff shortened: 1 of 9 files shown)


patrick-east:
Note that we will need to hold off merging this until the next OPA release (which should include the `--diagnostic-addr` feature).

internal/internal.go
```diff
@@ -169,11 +177,15 @@ func (p *envoyExtAuthzGrpcServer) listen() {
 		"enable-reflection": p.cfg.EnableReflection,
 	}).Info("Starting gRPC server.")
 
+	p.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateOK})
+
+	p.server.GetServiceInfo()
```
ashutosh-narkar, line 182: Why is this call being made ?

internal/internal_test.go
```diff
@@ -1077,6 +1080,73 @@ func TestGetParsedBody(t *testing.T) {
...
+}
+
+func assertPluginState(t *testing.T, m *plugins.Manager, expected plugins.State) {
+	t.Helper()
+	state, err := getPluginState(t, m)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if state != expected {
+		t.Fatalf("Exepected plugin state %v, got %v", expected, state)
```
ashutosh-narkar, line 1146: Nit: `Expected`

internal/internal.go
patrick-east: Oops, was just junk leftover from me exploring the API via intellisense. Removed it.

PR #188, opened 2020-07-07
allow using this with other plugins

srenatus:
When the definitions moved here live in a `main` package, they cannot be
imported by anyone else. Since the stuff they _use_ lives in an `internal`
package, you cannot define them in your own code.

Hence, there's no way to combine this plugin with another one, since you
cannot create the new cmd's code using the code here.

With this change, the same thing that happens in cmd/opa-istio-plugin/main.go
could happen in another main.go that also registers another plugin.

Exporting the `Factory` struct isn't necessarily the best option, happy to discuss alternatives 😃 Mostly, I wanted to raise the issue and present one way out here.


ashutosh-narkar:
@srenatus I'm curious if we have any particular use-case for this ?

srenatus:
We've been contemplating to add a custom rego builtin, and use it with this one.

But really, I think using any two plugins together would require this, in both plugins? Unless I'm missing something =)

tsandall:
Moving the factory into another package seems reasonable to me. It does not increase the API surface area too much since these APIs are already exported by OPA proper. LGTM.

srenatus:
Part of this is that I believe this plugin is used as a model for others -- like https://github.com/jonmclachlanatpurestorage/opa-oidc-plugin -- and it would generally help if plugins were "mixable"...

PR #194, opened 2020-08-11
internal: Updates to generated decision logs and non-OPA error wrapping

ashutosh-narkar:
This commit updates the server handler to create a decision ID and
a Metrics object on start-up that can be later included in the decision log.
This ensures that the decision log will have an ID and the server handler time
even in error scenarios.

This change also wraps non-OPA errors to ensure they are properly
serialized.

Fixes open-policy-agent/opa#2591, Fixes open-policy-agent/opa#2592,
Fixes open-policy-agent/opa#2593, Fixes open-policy-agent/opa#2594

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -9,6 +9,7 @@ import (
 	"crypto/rand"
 	"encoding/json"
 	"fmt"
+	"github.com/open-policy-agent/opa/topdown"
 	"io"
 	"net"
 	"net/url"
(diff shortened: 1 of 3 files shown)


internal/internal.go
```diff
@@ -189,11 +190,20 @@ func (p *envoyExtAuthzGrpcServer) listen() {
 
 func (p *envoyExtAuthzGrpcServer) Check(ctx ctx.Context, req *ext_authz.CheckRequest) (resp *ext_authz.CheckResponse, err error) {
 	start := time.Now()
+
+	result := evalResult{}
+	result.metrics = metrics.New()
+	result.metrics.Timer(metrics.ServerHandler).Start()
+	result.decisionID, err = uuid4()
+	if err != nil {
+		return nil, err
```
tsandall, line 199: Nit: I think case warrants a log message (at error level). If the server is unable to generate a UUID something is quite broken and it would be useful to see it in the logs. We can't send a decision log (because the ID can't be generated but at least someone looking at the console logs would be able to see the problem.)

tsandall: I guess it's possible that the gRPC server will log an error for us? I'm not sure how that works.

internal/internal.go
```diff
@@ -429,7 +429,16 @@ func (p *envoyExtAuthzGrpcServer) log(ctx context.Context, input interface{}, re
 		info.Metrics = result.metrics
 	}
 
-	if err == nil {
```
tsandall, line 432: Nit: the `if result != nil` check above can be removed since the `Check` function always passes a result to the `log` function now.

internal/internal.go
ashutosh-narkar: Yes the gRPC server will report this error.

internal/internal.go
ashutosh-narkar: Removed

internal/internal.go
tsandall: It'll log it? Can we rely on that?

ashutosh-narkar: I've seen it fail with the error reported. I'll add a log as well just in case.

PR #199, opened 2020-09-24
Fix port name in examples/istio/quick_start.yaml

skomma:
Currently, a Service resource within `examples/istio/quick_start.yaml` defines port 443 named "**http**".
But this configuration makes a situation which a pod in a mesh cannot connect to HTTPS endpoints.
(ex. istio/istio#14520)

This PR changes the port name to "https".

### DIFF
--- a/examples/istio/quick_start.yaml
+++ b/examples/istio/quick_start.yaml
@@ -149,7 +149,7 @@ metadata:
 spec:
   ports:
     - port: 443
-      name: http
+      name: https
   selector:
     app: admission-controller
 ---


ashutosh-narkar:
@skomma The change looks good. It would be helpful if you could elaborate how this change is related to the pods not being able to connect to https endpoints. That wasn't very clear from the issue you linked to.

skomma:
@ashutosh-narkar Thank you for your response.

> That wasn't very clear from the issue you linked to.

Sorry, but I cannot understand why this configuration breaks an SSL connection to the external services.
I hit the problem like [this post](https://stackoverflow.com/a/57641076).
The issue istio/istio#14520 was found in this post.

The another issue istio/istio#14264 that's encountering a similar situation might be more helpful.

ashutosh-narkar (approved):
LGTM

PR #200, opened 2020-09-29
docker: switch to static:nonroot variant of gcr.io/distroless

srenatus:
From online docs, adding the tag should be all that's needed.

With `docker run`, it looks like it was working:

    > opa.runtime()
    {
      "commit": "c1b006f3-dirty",
      "config": {},
      "env": {
        "HOME": "/home/nonroot",
        "HOSTNAME": "86248ee2c14a",
        "PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
        "SSL_CERT_FILE": "/etc/ssl/certs/ca-certificates.crt",
        "TERM": "xterm"
      },
      "version": "0.23.2-envoy-1"
    }

Since we don't need root, we shouldn't use it 😃 

👉 https://github.com/GoogleContainerTools/distroless/issues/235 This is where it says that adding the tag should be sufficient.


Dockerfile
```diff
@@ -2,9 +2,9 @@
 # Use of this source code is governed by an Apache2
 # license that can be found in the LICENSE file.
 
-FROM gcr.io/distroless/base
+FROM gcr.io/distroless/static:nonroot
```
patrick-east, line 5: Out of curiosity, is there a difference between the `static` and `base` image types? I looked around on the distroless readme and didn't spot any mention of the static ones.

srenatus:
https://github.com/GoogleContainerTools/distroless/tree/master/base#image-contents
it's libc, openssl and libssl. Haven't checked how much smaller the images
become yet. :)

tsandall:
Circling back to this...

We've had to change the non-Envoy versions of the openpolicyagent/opa image to use `gcr.io/distroless/cc` because of the embedded wasm runtime feature. We'll probably have to do the same for this version of the image...because of this, should we just close this PR for now?

srenatus:
:+1:

PR #201, opened 2020-09-29
golang: 1.13.7 -> 1.15.2

srenatus:
Adapted tests to work with that golang version.

Also ran `go mod tidy`.

PR #202, opened 2020-10-01
support basic gRPC payload decoding

pnnferreira:
with the new [functionality from envoy](https://github.com/envoyproxy/envoy/pull/13131) to "pack_as_bytes", opa can decode the raw body into the parsed body to make policy decisions.


internal/internal.go
```diff
@@ -9,27 +9,34 @@ import (
...
+	ext_authz "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	ext_type "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/genproto/googleapis/rpc/code"
 	rpc_status "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
 
+	proto "github.com/golang/protobuf/proto"
```
srenatus, line 38: [nit] it's already called proto, so we should be able to do:
```suggestion
	"github.com/golang/protobuf/proto"
```

internal/internal.go
```diff
@@ -133,6 +140,7 @@ type Config struct {
 	DryRun           bool   `json:"dry-run"`
 	EnableReflection bool   `json:"enable-reflection"`
 	parsedQuery      ast.Body
+	ProtoDescriptor  string `json:"proto_descriptor"`
```
srenatus, line 143: The new configurable should be documented somewhere 😃

internal/internal.go
```diff
@@ -612,39 +620,121 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+
+			if len(rawbody) <= 0 {
+				return nil, false, fmt.Errorf("invalid raw body")
+			}
+			if len(parsedPath) <= 1 {
+				return nil, false, fmt.Errorf("invalid parsed path")
+			}
+
+			if protoDescriptor == "" {
+				return nil, false, fmt.Errorf("ProtoDescriptor path not defined")
```
srenatus, line 662: ❓Should this be an error? Imagine I'm using this and I don't care about authorizing gRPC at all. I wouldn't set this field. If a gRPC request was sent to opa-envoy-plugin then, should it not just ignore the body?

I'm thinking something along the lines of 
```go
		} else if strings.Contains(val, "application/grpc") {
			if protoDescriptor == "" {
				return nil, false, nil
			}
			// ...
}
```
Curious what you, and others, think about this case 😃

internal/internal.go
```diff
@@ -612,39 +620,121 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+	}
+
+	fd, err := desc.CreateFileDescriptorFromSet(&fileSet)
+	if err != nil {
+		return err
+	}
+
+	inputType := ""
+	packageName := fd.GetPackage()
+	pathServiceRaw := string(parsedPath[0].(string))
```
srenatus, line 700: Hmm I think this would be enough:
```suggestion
	pathServiceRaw := parsedPath[0].(string)
```

internal/internal.go
```diff
@@ -612,39 +620,121 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+	pathService := strings.Replace(pathServiceRaw, packageName+".", "", 1)
+
+	for _, v := range fd.GetServices() {
+		if v.GetName() == pathService {
+			for _, z := range v.GetMethods() {
+				if z.GetName() == parsedPath[1] {
```
srenatus, line 706: Does it not care about `parsedPath[1]` being `interface{}` and not `string`? 🤔

internal/internal_test.go
```diff
@@ -1272,12 +1274,129 @@ func TestGetParsedBody(t *testing.T) {
...
+
+	expectedObject2 := map[string]interface{}{}
+	expectedObject2["SeverityNumber"] = "SecNumber"
+	expectedObject2["SeverityText"] = "ERROR"
+
+	expectedObject := map[string]interface{}{}
+	expectedObject["Data"] = expectedObject1
+	expectedObject["Metadata"] = expectedObject2
+
+	//grpcserver := testAuthzServer(&testPlugin{}, false)
```
srenatus, line 1354: These should be cleaned up 🧹

internal/internal_test.go
```diff
@@ -1272,12 +1274,129 @@ func TestGetParsedBody(t *testing.T) {
...
+
+	//grpcserverWithConfig := testAuthzServer(&testPlugin{}, false)
+	//grpcserverWithConfig.cfg.ProtoDescriptor = `../test/files/example.pb`
+
+	protoDescriptorPath := `../test/files/example.pb`
+
+	tests := map[string]struct {
+		input               *ext_authz.CheckRequest
+		want                interface{}
+		isBodyTruncated     bool
```
srenatus, line 1364: I don't think we currently support proper detection of a truncated protobuf body, do we? This could be a follow-up, I suppose.

test/files/Example.proto
```diff
@@ -0,0 +1,65 @@
+syntax = "proto3";
```
srenatus, line 1: This looks like a real-world-example. That in itself is not a problem, but it would be nice to have an example file that specifically covers different cases:

- enums
- nested messages
- maybe an imported input type, like `google.protobuf.Empty`?

test/files/GenerateProto.md
```diff
@@ -0,0 +1,3 @@
+Folder: TestGrpc
+
+protoc --proto_path=TestGrpc --go_out=testGrpc --go_opt=paths=source_relative TestGrpc/Example.proto --descriptor_set_out=TestGrpc/example.pb --include_imports
```
srenatus, line 3: Can you make this a README or a Makefile? Also I believe the paths aren't correct at the moment. 🤔

internal/internal_test.go
```diff
@@ -1062,7 +1062,7 @@ func TestGetResponseHttpStatus(t *testing.T) {
 		t.Fatal("Expected error but got nil")
 	}
 
-	input["http_status"] = json.Number(301)
+	input["http_status"] = json.Number("1")
```
srenatus, line 1065: I suspect that this can disappear when rebasing this branch. It had been dealt with in #201.

srenatus (commented):
👏 🎉 Glad to see the PR opened! Added a few comments inline 👇

internal/internal.go
pnnferreira: it makes sense, its a way to turn off the the capability to decode the raw body

internal/internal_test.go
pnnferreira: currently i dont think so

pnnferreira:
@ashutosh-narkar Do you have any idea when be the feature will be added on a release?

srenatus:
Couldn't find the commit @ashutosh-narkar has been referring to (or having in mind), but 0.9.7 was released half a day ago (https://github.com/envoyproxy/go-control-plane/releases/tag/v0.9.7) and since it corresponds to what's in `master` for that repo, I'd assume the release is there now. @ashutosh-narkar Am I missing something? Can you point us to the code change there you've been thinking about, please? 😃

ashutosh-narkar:
`v0.9.7` is the release I was referring to before. @pnnferreira can you please update the PR to use the new release ? Thanks

pnnferreira:
i have updated the control plane version

internal/internal_test.go
```diff
@@ -1272,12 +1274,145 @@ func TestGetParsedBody(t *testing.T) {
...
+
+	expectedObject2 := map[string]interface{}{}
+	expectedObject2["SeverityNumber"] = "SecNumber"
+	expectedObject2["SeverityText"] = "ERROR"
+
+	expectedObject := map[string]interface{}{}
+	expectedObject["Data"] = expectedObject1
+	expectedObject["Metadata"] = expectedObject2
+
+	protoDescriptorExamplePath := `../test/files/example/example.pb`
```
srenatus, line 1369: The test failure is because the tests are run on linux: the filesystem is case-sensitive. The committed `example.pb` is in `test/files/Example/example.pb`, and referenced in the test code as `test/files/example/example.pb`. That works fine on case-insensitive FS (OSX, maybe Windows?) but fails in travis.

srenatus: Can you make the committed path lowercase, please? It's more common, I think.

pnnferreira: ok changed it. The tests are successful now

pnnferreira:
@ashutosh-narkar any obstacle to move forward with this PR?

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+			if protoDescriptor == "" {
+				return nil, false, nil
+			}
+
+			rawbody := req.GetAttributes().GetRequest().GetHttp().GetRawBody()
+
+			if len(rawbody) <= 0 {
+				return nil, false, fmt.Errorf("invalid raw body")
+			}
+			if len(parsedPath) <= 1 {
```
ashutosh-narkar, line 662: Why does the parsed path need to have a length of 2 ?

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+
+	fd, err := desc.CreateFileDescriptorFromSet(&fileSet)
+	if err != nil {
+		return err
+	}
+
+	inputType := ""
+	packageName := fd.GetPackage()
+	pathServiceRaw := parsedPath[0].(string)
+	pathService := strings.Replace(pathServiceRaw, packageName+".", "", 1)
```
ashutosh-narkar, line 702: It's not clear to me what's happening here. Some comments may help.

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+			for _, z := range v.GetMethods() {
+				if z.GetName() == parsedPath[1] {
+					inputType = z.GetInputType().GetName()
+				}
+			}
+		}
+	}
+
+	if packageName == "" || inputType == "" {
+		return fmt.Errorf("InputType not defined")
```
ashutosh-narkar, line 715: Separating these checks with the right error message will be helpful.

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
 
 	return data, false, nil
 }
 
+func getGRPCBody(in []byte, parsedPath []interface{}, data interface{}, protoDescriptor string) error {
+
+	// the first 5 bytes are part of gRPC framing. We need to remove them to be able to parse
+	//https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
+
+	in = in[5:]
```
ashutosh-narkar, line 681: Don't you need to check the size here ?

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+	}
+
+	jsonBody, err := json.Marshal(message)
+	if err != nil {
+		return err
+	}
+
+	if err := util.Unmarshal([]byte(jsonBody), &data); err != nil {
+		return err
+	}
```
ashutosh-narkar, line 735: Can this be avoided ?

ashutosh-narkar (commented):
@pnnferreira this seems like a good start. I had some comments about the implementation and some nits. Also we should look into [Go protocol buffers APIv2](https://pkg.go.dev/google.golang.org/protobuf?tab=overview). It may cut down on the dependencies.

go.mod
```diff
@@ -3,23 +3,28 @@ module github.com/open-policy-agent/opa-envoy-plugin
...
+	github.com/google/go-cmp v0.5.1 // indirect
 	github.com/gorilla/mux v1.7.3 // indirect
+	github.com/jhump/protoreflect v1.7.0
+	github.com/juju/testing v0.0.0-20200923013621-75df6121fbb0 // indirect
 	github.com/kr/pretty v0.1.0 // indirect
 	github.com/mattn/go-runewidth v0.0.4 // indirect
 	github.com/open-policy-agent/opa v0.23.2
 	github.com/peterh/liner v1.1.0 // indirect
 	github.com/pkg/errors v0.8.1
+	github.com/pnnferreira/grpcopa v0.0.0-20200924172237-988227547a93 // indirect
```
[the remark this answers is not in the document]
srenatus: I guess `go mod tidy` would clean this up... 🤔

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
 		}
 	}
 
 	return data, false, nil
 }
 
+func getGRPCBody(in []byte, parsedPath []interface{}, data interface{}, protoDescriptor string) error {
+
+	// the first 5 bytes are part of gRPC framing. We need to remove them to be able to parse
+	//https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
```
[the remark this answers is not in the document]
srenatus: I think we should be able to use this instead: https://pkg.go.dev/github.com/matttproud/golang_protobuf_extensions/pbutil#ReadDelimited

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
 			}
+		} else if strings.Contains(val, "application/grpc") {
+
+			if protoDescriptor == "" {
+				return nil, false, nil
+			}
+
+			rawbody := req.GetAttributes().GetRequest().GetHttp().GetRawBody()
+
+			if len(rawbody) <= 0 {
```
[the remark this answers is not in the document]
pnnferreira: the idea is that when we do not set the protoDescriptor this feature is turned off, that is why we return nil on line 654.
We assume that if the we set the path it is to use this feature, and we must have a raw body, if we don't have one i think we should warn the user. 
Anyway i could change it if you want to

internal/internal.go
pnnferreira: Because we need to know the input type to decode. To find the input type to decode we need have the package + method, 
that information its present in parsedpath[1]

internal/internal.go
pnnferreira: //To get the service.
//The package and service are in the parsedpath[0] (Example.Test.GRPC.ProtoServiceIExampleApplication), we need to get the service (ProtoServiceIExampleApplication) to find the methods available

internal/internal.go
```diff
@@ -612,39 +620,123 @@ func getParsedPathAndQuery(req *ext_authz.CheckRequest) ([]interface{}, map[stri
...
+		return fmt.Errorf("InputType not defined")
+	}
+
+	messageName := fmt.Sprintf("%s.%s", packageName, inputType)
+
+	msgDesc := fd.FindMessage(messageName)
+
+	message := dynamic.NewMessage(msgDesc)
+
+	if err := proto.Unmarshal(in, message); err != nil {
```
srenatus, line 724: I believe instead of skipping the first few bytes ([above](https://github.com/open-policy-agent/opa-envoy-plugin/pull/202/files#diff-67213f39b84a013e7db04664cd3ad943b7a237d97b9c6e862d10806ac27def28R681)), we might be able to do something like this _here_:

```suggestion
	if _, err := pbutil.ReadDelimited(bytes.NewReader(in), message); err != nil
```

srenatus:
> Also we should look into Go protocol buffers APIv2. It may cut down on the dependencies.

@ashutosh-narkar I agree, and I volunteer dealing with this in a follow-up PR. Let's get this in first. 😃

internal/internal.go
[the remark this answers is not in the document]
srenatus: :bulb: I was wrong! 5 is alright, it's documented here:

> The repeated sequence of Length-Prefixed-Message items is delivered in DATA frames
> 
> Length-Prefixed-Message → Compressed-Flag Message-Length Message
> Compressed-Flag → 0 / 1 # encoded as 1 byte unsigned integer
> Message-Length → {length of Message} # encoded as 4 byte unsigned integer (big endian)
> Message → *{binary octet}

So 1 byte is Compressed-Flag, the next 4 are the Message-Length.

(The ReadDelimited method from pbutil is for situations when the message is prefixed with a varint size, which isn't the case of gRPC's mapping to HTTP2.)

internal/internal.go
pnnferreira: how?

srenatus:
I think it would be good if could get this merged and go on from there.

Right now, what's missing, I believe, is:
- [ ] document the limitations: no compression, only a single message; also only protobuf (gRPC could be used with other encodings)
- [ ] rebase, squash the branch

In follow-up PRs, we can deal with
- decompression of request bodies, or logging if a compressed on was sent
- dealing with multiple messages, perhaps?
- eventually switching to APIv2 of protobuf

internal/internal.go
srenatus: I was wrong here, see https://github.com/open-policy-agent/opa-envoy-plugin/pull/202#discussion_r506349361

srenatus:
We've discussed this PR in slack and @pnnferreira thought it would be a good idea if took over.

So, I'm going to pick this up soon. 

@pnnferreira I could re-use *this* PR if you made me a collaborator on your fork, https://github.com/pnnferreira/opa-envoy-plugin, please 😃

pnnferreira:
@srenatus as we discussed on slack i agree. 
I have already made me you a collaborator on your fork. thanks 👍

srenatus:
Alright, I've started working on this. It looks like I cannot switch it back to draft, but I'll let y'all know when to have another look. 😎

internal/internal.go
```diff
@@ -103,11 +109,10 @@ func Validate(m *plugins.Manager, bs []byte) (*Config, error) {
 func New(m *plugins.Manager, cfg *Config) plugins.Plugin {
 
 	plugin := &envoyExtAuthzGrpcServer{
-		manager:                m,
-		cfg:                    *cfg,
-		server:                 grpc.NewServer(),
-		preparedQueryDoOnce:    new(sync.Once),
-		interQueryBuiltinCache: iCache.NewInterQueryCache(m.InterQueryBuiltinCacheConfig()),
```
srenatus, line 110: TODO merge/rebase messup by me

ashutosh-narkar:
@srenatus I've moved the PR to draft mode. Feel free to switch it back. Thanks !

srenatus:
I've cleaned up a few things here:

- parse the protoset once only (in plugin initialization)
- not building the golang pb code, we don't need it
- using a combined protoset for multiple services (like envoy does it in grpc_json transcoding)
- using protobuf APIv2

Remaining TODOs:

- [ ] see if e2e tests can be expanded to cover this
- [ ] play around with the plugin and envoy some more, look into edge cases (streaming, compression)

srenatus:
Now, we're dealing properly with edge cases (empty messages, client compression, truncated payloads).

Next, I'll add an `examples/grpc/` folder, with a README on how to generate the protoset, and some example calls.
The stuff from that folder can then help extending the e2e tests we currently have. Also, I've probably broken the existing ones, since we're going to ext auth v3 here.

💭 Thinking about it, going to v3 might make more sense in a separate PR, even if it blocks this one. I'd like to avoid this one becoming a hairy mud ball mega-PR.

tsandall:
>Next, I'll add an examples/grpc/ folder, with a README on how to generate the protoset, and some example calls.
The stuff from that folder can then help extending the e2e tests we currently have. Also, I've probably broken the existing ones, since we're going to ext auth v3 here.

>Thinking about it, going to v3 might make more sense in a separate PR, even if it blocks this one. I'd like to avoid this one becoming a hairy mud ball mega-PR.

Are there backwards compatibility considerations here? What is the minimum version of Envoy that will work? We don't necessarily have to be too strict here but it would be good to know. If we break compatibility with older versions of Envoy, I'm sure there will be users who run into this.

srenatus:
> Are there backwards compatibility considerations here? What is the minimum version of Envoy that will work? We don't necessarily have to be too strict here but it would be good to know. If we break compatibility with older versions of Envoy, I'm sure there will be users who run into this.

From the discussion of https://github.com/open-policy-agent/opa/issues/2798, it looks like the support for v2 will be dropped next month. But I can follow-up on it. We could support both gRPC APIs (v2 and v3) in one service, if there's a considerable need for it.

I'm afraid this is not 100% accurate,
> Even after incorporating the V3 API the plugin itself would still be backward compatible so I'm not too worried about breaking anyone

...after switching to v3, the service will just not recognize v2 requests anymore.

All the more reason to detangle this from the gRPC body feature.

srenatus:
This isn't left to rot, I'll pick it up when #217, its spinoff, is merged..

srenatus:
With #217 out of the way, I'm picking this up again. 

Rebase is done, now adding something to our e2e test.

internal/internal.go
```diff
@@ -695,36 +725,131 @@ func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{},
...
 				return nil, false, err
 			}
+		} else if strings.Contains(val, "application/grpc") {
+
+			if protoSet == nil {
+				return nil, false, nil
+			}
+
+			if len(rawBody) == 0 {
+				return nil, false, fmt.Errorf("invalid raw body")
```
srenatus, line 759: 🚧 This has to change: With the other corner cases, we've left it to the policy author to decide what to do -- if the body couldn't be parsed, or if it was truncated; **this case here** would be hit if opa-envoy-plugin was configured to parse the protobuf bodies but didn't get one. It should behave the same, `return nil, false, nil`.

internal/internal.go
```diff
@@ -695,36 +724,131 @@ func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{},
...
+func getGRPCBody(in []byte, parsedPath []interface{}, data interface{}, files *protoregistry.Files) (found, truncated bool, _ error) {
+
+	// the first 5 bytes are part of gRPC framing. We need to remove them to be able to parse
+	// https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
+
+	if len(in) < 5 {
+		return false, false, fmt.Errorf("less than 5 bytes")
+	}
+
+	// Can be 0 or 1, 1 indicates thet the payload is compressed.
```
ashutosh-narkar, line 789: Nit: `thet `

internal/internal.go
```diff
@@ -695,36 +724,131 @@ func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{},
...
+	// The method could be looked up in the request headers, and the
+	// request decompressed; but for now, let's skip it.
+	if in[0] != 0 {
```
ashutosh-narkar, line 792: We are simply checking here if the payload is compressed and for now we won't support that.

internal/internal.go
```diff
@@ -695,36 +724,131 @@ func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{},
...
+		return false, false, nil
+	}
+
+	// Note: we're only reading one message, this is the first message's size
+	size := binary.BigEndian.Uint32(in[1:5])
+	if int(size) > len(in)-5 {
+		return false, true, nil // truncated body
+	}
+	in = in[5 : size+5]
```
ashutosh-narkar, line 801: Is everything from the 6th byte the actual input ?

internal/internal.go
```diff
@@ -695,36 +724,131 @@ func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{},
...
+
+	// Note: we've already checked that len(path)>=2
+	svc, err := findService(parsedPath[0].(string), files)
+	if err != nil {
+		return false, false, nil
+	}
```
ashutosh-narkar, line 807: Why don't we return the error ?

internal/internal.go
```diff
@@ -695,36 +724,131 @@ func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{},
...
+	msgDesc, err := findMessageInputDesc(parsedPath[1].(string), svc)
+	if err != nil {
+		return false, false, nil
+	}
```
ashutosh-narkar, line 811: Why don't we return the error ? Is it because of the unsupported case ?

internal/internal.go
```diff
@@ -695,36 +724,131 @@ func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{},
...
+	if !ok {
+		return nil, fmt.Errorf("could not find service descriptor for path %q", path)
+	}
+	return svcDesc, nil
+}
+
+func findMessageInputDesc(name string, svc protoreflect.ServiceDescriptor) (protoreflect.MessageDescriptor, error) {
+	if method := svc.Methods().ByName(protoreflect.Name(name)); method != nil {
+		if method.IsStreamingClient() {
+			return nil, fmt.Errorf("method %s is a streaming client method, unsupported", method.Name())
```
ashutosh-narkar, line 845: Nit: Let's reword the error message a bit. For example, `streaming client method foo not supported`

internal/internal_test.go
```diff
@@ -992,6 +993,52 @@ func TestConfigInvalid(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	in := `{"proto-descriptor": "this/does/not/exist"}`
+	_, err = Validate(m, []byte(in))
+	if err == nil {
+		t.Error("expected err, got nil")
+	}
+}
+
```
ashutosh-narkar, line 1041: `TestConfigWithNonexistentProtoDescriptor ` got covered in `TestConfigWithInvalidProtoDescriptor `.

Nit: We could combine `TestConfigWithValidProtoDescriptor ` and `TestConfigWithInvalidProtoDescriptor` into a single table driven test.

internal/internal_test.go
```diff
@@ -1487,13 +1538,238 @@ func TestGetParsedBody(t *testing.T) {
...
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			headers := tc.input.GetAttributes().GetRequest().GetHttp().GetHeaders()
+			body := tc.input.GetAttributes().GetRequest().GetHttp().GetBody()
+			rawBody := tc.input.GetAttributes().GetRequest().GetHttp().GetRawBody()
+			path := tc.input.GetAttributes().GetRequest().GetHttp().GetPath()
+
+			parsedPath, parsedQuery, err := getParsedPathAndQuery(path)
+			if err != nil {
+				t.Fatalf("expected error: %v, parsedQuery: %v", tc.err, parsedQuery)
```
ashutosh-narkar, line 1752: Do we expect an error here ?

internal/internal.go
srenatus: The structure is `<size><payload><size><payload>...` if there are multiple messages (i.e. a streaming gRPC method). So, we only support _one message_ for now, and only read that one (if we've been given enough bytes).

internal/internal.go
srenatus: If we don't bail on these, the policy code can decide what to do with a payload that OPA didn't know how to parse.

They could still forbid any request that doesn't have a `input.parsed_body`; but we'd allow them to do other things.

I think it's conceivable, for example, that some services (like health checks) aren't included in the proto descriptor set, and the `input.parsed_path` is used to allow them. If we returned the error here, that wouldn't be possible, every gRPC request ever made would have to be part of our proto descriptor set.

internal/internal.go
srenatus: Same as with `findService`, I'd say.

internal/internal_test.go
srenatus: Good catch 😄

.github/workflows/post-merge.yaml
```diff
@@ -86,3 +86,26 @@ jobs:
...
+
+      - name: Build docker image
+        run: make image tag-latest
+
+      - name: Build testsrv docker image
+        run: make testsrv-image
+        working-directory: examples/grpc
+
+      - name: Run test
+        run: make test-setup test
```
srenatus, line 105: I believe this _could be_ flakey. `test-setup` run `docker-compose up -d`, and the tests start calling the service; there's a chance that envoy isn't ready yet.

However, I haven't been able to trigger this so far locally,
```interactive
$ docker-compose up -d; grpcurl -plaintext -protoset testsrv.pb 127.0.0.1:51051 test.KitchenSink/Ping
Creating network "grpc_default" with the default driver
Creating grpc_envoy_1     ... done
Creating grpc_testsrv_1   ... done
Creating grpc_opa-envoy_1 ... done
{

}
$
```
and on github, we're more than likely to pull down the grpcurl image anyways, given envoy plenty of time. While I'm not happy with this arrangement, let's try to roll with it. We'll deal with it if it becomes a problem.

internal/internal.go
srenatus: Ultimately, I'd love some user feedback on this. 😃

srenatus:
ℹ️ We're rebuilding the testsrv image although it's very unlikely to ever change. Since it's below one minute, I've decided that it might be good enough for now. Future alternatives: push the image to dockerhub, or use grpcbin instead (but they're lacking some WKT, https://github.com/moul/pb/blob/master/grpcbin/grpcbin.proto#L67)

.github/workflows/pull-request.yaml
```diff
@@ -66,3 +66,26 @@ jobs:
...
+        working-directory: examples/grpc
+
+      - name: Run test
+        run: make test-setup test
+        working-directory: examples/grpc
+
+      - name: Run test log dump and cleanup
+        run: make test-teardown
+        if: ${{ always() }}
+        working-directory: examples/grpc
```
ashutosh-narkar, line 91: It would be great if we could find a way to avoid all the duplication. We could create a make target but probably won't look that great in the UI.

README.md
```diff
@@ -175,6 +177,14 @@ The `enable-reflection` parameter registers the Envoy External Authorization gRP
 server reflection, a command line tool such as [grpcurl](https://github.com/fullstorydev/grpcurl) can be used to invoke
 RPC methods on the gRPC server. See [gRPC Server Reflection Usage](#grpc-server-reflection-usage) section for more details.
 
+Providing a file containing a protobuf descriptor set allows the plugin to decode gRPC message payloads.
+So far, only unary methods using uncompressed protobuf-encoded payloads are supported.
+The protoset can be generated using `protoc`, e.g. `protoc --descriptor_set_out=protoset.pb --include_imports`.
+
+Note that gRPC message payload decoding is only available [using the v3 API](#envoy-xds-v2-and-v2).
+See [`examples/grpc`](examples/grpc) an example setup using Envoy, a gRPC service, and opa-envoy-plugin examining the
```
ashutosh-narkar, line 185: Nit: `for` an example setup using Envoy ...

examples/grpc/docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+    volumes:
+      - ./envoy.yaml:/etc/envoy/envoy.yaml
+  opa-envoy:
+    image: openpolicyagent/opa:latest-istio
+    ports:
+      - "9191:9191"
+    command:
+      - run
+      - --server
+#      - --log-level=debug
```
ashutosh-narkar, line 17: Nit: remove

examples/grpc/docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+services:
+  envoy:
+    image: envoyproxy/envoy:v1.16-latest
+    ports:
+      - "9901:9901"
+      - "51051:51051"
+    volumes:
+      - ./envoy.yaml:/etc/envoy/envoy.yaml
+  opa-envoy:
+    image: openpolicyagent/opa:latest-istio
```
ashutosh-narkar, line 11: Nit: `latest-envoy`

internal/internal.go
ashutosh-narkar: Yes, we should at least log these errors. It would help to debug any future issues.

.github/workflows/pull-request.yaml
srenatus: Totally agreed! I think github actions have a feature for that, the one or two steps that are specific to each variant (PR check, post-merge) can probably be `if:`'ed accordingly.

srenatus:
ℹ️  This would be the kind of (debug) log we'd get when a gRPC request was passed to opa-envoy-plugin that we couldn't make sense of given the provided descriptor set:

```
{"decision-id":"00347a74-7f36-4c21-bde0-23a8cdd3524e","err":"proto: not found","level":"debug","msg":"could not find service","time":"2020-12-03T09:17:09Z"}
```

srenatus:
Looks like this has bitrot again. I'll rebase.

internal/internal_test.go
```diff
@@ -1430,18 +1457,18 @@ func TestGetParsedBody(t *testing.T) {
...
 	tests := map[string]struct {
-		input           *ext_authz.CheckRequest
-		want            interface{}
-		isBodyTruncated bool
-		err             error
+		input               *ext_authz.CheckRequest
+		want                interface{}
+		isBodyTruncated     bool
+		err                 error
+		protoDescriptorPath string
```
ashutosh-narkar, line 1471: Nit: Is `protoDescriptorPath ` used in this test ?

srenatus: Done ✔️

ashutosh-narkar (approved):
LGTM.

ashutosh-narkar:
@srenatus can you please fix the conflict and then we can merge !

srenatus:
@ashutosh-narkar done! :)

srenatus:
@pnnferreira Look at that! Finally 🎊 Thanks for all your help and effort getting this off the ground.

pnnferreira:
Great news  :smiley: Thanks for all the help and for finishing this :clap:

PR #204, opened 2020-10-14
deps: update opa to v0.24.0

srenatus:
Trying some stuff to make #203 work.

PR #205, opened 2020-10-14
update-opa-version.sh: remove workaround

srenatus:
Tried this out in #204. Seems to do fine 😃


ashutosh-narkar:
Thanks @srenatus ! I tried this out and it seems to work as well.

PR #208, opened 2020-10-28
Makefile: Use the plugin version to build releases

ashutosh-narkar:
Earlier we used only the OPA version to create a GH release and
build the release binaries. We did not create a new GH release when
changes we made to the plugin itself.

This change updates the release target to use the plugin version
instead of only the OPA version to checkout and build binaries.
This will allow us to create releases for the plugin revisions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ endif
 
 RELEASE_BUILD_IMAGE := golang:$(GOVERSION)
 
-RELEASE_DIR ?= _release/$(VERSION_OPA)
+RELEASE_DIR ?= _release/$(VERSION)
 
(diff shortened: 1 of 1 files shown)


tsandall:
LGTM

Q: what are the steps required to create a release for the plugin? Is it just a matter of creating the tag for the release commit?

tsandall (approved):

ashutosh-narkar:
> LGTM
> 
> Q: what are the steps required to create a release for the plugin? Is it just a matter of creating the tag for the release commit?

Yes. Simply create a release with the appropriate tag and the GHA will create the binaries for that version.

PR #210, opened 2020-10-29
.github: Change fetch-depth in the GHA checkout step

ashutosh-narkar:
Earlier the checkout step of the post-tag workflow
used the default fetch-depth=1 which would only fetch a
single commit. This would result in an incorrect calculation
of the number of commits from the last OPA revendoring.

This change sets fetch-depth=0 to fetch all history to correct
this issue.


srenatus (approved):
LGTM

PR #211, opened 2020-11-06
internal: Handle errors from OPA eval cancellation

ashutosh-narkar:
When the context is cancelled or deadline is reached
durign query evaluation, the error message returned is
"caller cancelled query execution". This message is
not very informative from the pov of the plugin user.
This change updates the plugin to handle eval cancellation
errors and returns a more helpful error message.

This change also checks if the context is cancelled or
deadline is reached even before policy eval has started
which may happen in scenarios where the server is overloaded and
has pending requests in its queue and returns an appropriate
error message in such cases.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -9,6 +9,7 @@ import (
 	"crypto/rand"
 	"encoding/json"
 	"fmt"
+	"github.com/open-policy-agent/opa/topdown"
 	"io"
 	"net"
 	"net/url"
@@ -34,7 +35,6 @@ import (
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/server"
 	"github.com/open-policy-agent/opa/storage"
-	"github.com/open-policy-agent/opa/topdown"
 	iCache "github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/open-policy-agent/opa/util"
 )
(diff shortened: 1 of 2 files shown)


internal/internal.go
```diff
@@ -190,12 +190,25 @@ func (p *envoyExtAuthzGrpcServer) listen() {
...
+
+	// Check if context is cancelled or deadline is reached.
+	exit := make(chan struct{})
+	defer close(exit)
+	go waitForDone(ctx, exit, func() {
+		err = fmt.Errorf("context deadline reached due to server overload")
+	})
+	if err != nil {
+		return nil, err
+	}
```
patrick-east, line 202: Can we check for `ctx.Err()` rather than using the done channel directly? If it returns anything except nil we know that done is complete.

ashutosh-narkar: Updated

patrick-east: Ah sry, i guess i meant not using the background thread and instead doing something more simple like:

```
if ctx.Err() != nil {
	return nil, errors.Wrap(ctx.Err(), "context deadline reached due to server overload")
}
```

Was the intention to have that background thread waiting to catch cancellations during evaluation too?

ashutosh-narkar: You're right. We don't need to do that.

internal/internal.go
```diff
@@ -190,12 +190,19 @@ func (p *envoyExtAuthzGrpcServer) listen() {
 }
 
 func (p *envoyExtAuthzGrpcServer) Check(ctx context.Context, req *ext_authz.CheckRequest) (resp *ext_authz.CheckResponse, err error) {
+	if ctx.Err() != nil {
+		err = errors.Wrap(ctx.Err(), "server overload error")
+		return nil, err
+	}
+
```
tsandall, line 197: I think we should move this down to L225 so that we see this error in the decision log (as is, it would not be logged.) Also, in terms of wording, I'm not sure what the final message looks like (since we're using `errors.Wrap` here) but I'd like it to indicate that the check request was timed out before query execution started.

ashutosh-narkar: Updated. The final error message looks something like  -> `check request timed out before query execution: context deadline exceeded`.

tsandall: LGTM!

PR #216, opened 2020-11-19
internal: Handle array of response headers

keyolk:
Hi there,
I put simple code that handle array of response headers.
It will take multiple headers which has same key.

I found similar requests from the issue below
https://github.com/open-policy-agent/opa/issues/2628

Signed-off-by: keyolk <chanhun.jeong@navercorp.com>


internal/internal.go
```diff
@@ -492,35 +493,57 @@ func getResponseStatus(result map[string]interface{}) (int32, error) {
...
+				responseHeaders = append(responseHeaders, responseHeadersToAppend...)
+			} else {
+				return nil, fmt.Errorf("type assertion error")
+			}
 		}
 
-		headerValueOption := &ext_core.HeaderValueOption{
-			Header: headerValue,
+	case map[string]interface{}:
+		responseHeadersToUse, err := takeResponseHeaders(val.(map[string]interface{}))
```
srenatus, line 537: Let's do this a little differently:
```go
switch val := val.(type) {
case []interface{}:
	for _, vval := range val { // val is []interface{} here, no need to cast again
	// ...
case map[string]interface{}:
	responseHeadersToUse, err := takeResponseHeaders(val) // likewise
	// ...
}
```

internal/internal_test.go
```diff
@@ -1114,6 +1115,23 @@ func TestGetResponseHeaders(t *testing.T) {
...
 	}
+
+	testHeaders = make(map[string]interface{})
+	testHeaders["foo"] = "bar"
+	testHeaders2 := make(map[string]interface{})
+	testHeaders2["foo"] = "baz"
+	testHeadersSlices := make([]interface{}, 0)
+	testHeadersSlices = append(testHeadersSlices, testHeaders)
+	testHeadersSlices = append(testHeadersSlices, testHeaders2)
+	input["headers"] = testHeadersSlices
```
srenatus, line 1126: It's probably just a matter of preference -- what do you think about this? 

```suggestion
	input["headers"] = []interface{}{
		map[string]interface{}{
			"foo": "bar",
		},
		map[string]interface{}{
			"foo": "baz",
		},
	}
```

I find it slightly easier to read.

internal/internal_test.go
```diff
@@ -1114,6 +1115,23 @@ func TestGetResponseHeaders(t *testing.T) {
...
+
+	result, err = getResponseHeaders(input)
+	if err != nil {
+		t.Fatalf("Expected no error but got %v", err)
+	}
+	if len(result) != 2 {
+		t.Fatalf("Expected two header but got %v", len(result))
+	}
```
srenatus, line 1134: Would you think it be useful to have more concrete assertions on what is returned there? 🤔

internal/internal.go
```diff
@@ -492,35 +493,57 @@ func getResponseStatus(result map[string]interface{}) (int32, error) {
...
+		for _, vval := range val.([]interface{}) {
+			if headers, ok := vval.(map[string]interface{}); ok {
+				responseHeadersToAppend, err := takeResponseHeaders(headers)
+				if err != nil {
+					return nil, err
+				}
+				responseHeaders = append(responseHeaders, responseHeadersToAppend...)
+			} else {
+				return nil, fmt.Errorf("type assertion error")
+			}
```
srenatus, line 533: We could try to avoid a level of indentation here

```suggestion
			headers, ok := vval.(map[string]interface{})
			if !ok {
				return nil, fmt.Errorf("type assertion error")
			}
			
			responseHeadersToAppend, err := takeResponseHeaders(headers)
			if err != nil {
				return nil, err
			}
			responseHeaders = append(responseHeaders, responseHeadersToAppend...)
```

It's a matter of taste, I suppose. 💭

srenatus (commented):
Thank you for picking this up! It's a good improvement. 😃 

Some nitpicks aside (see inline comments), would you mind adding a short paragraph to the docs on object responses? It'll be most useful to everyone if this nice feature is surfaced there. 🎉

internal/internal.go
ashutosh-narkar: Thanks for adding this @keyolk ! It would be useful if you could add a policy example in [this](https://github.com/open-policy-agent/opa-envoy-plugin#example-policy-with-object-response) section with some text (like Stephan mentioned) and also add a policy example.

internal/internal_test.go
keyolk: 🤔 checking duplicated headers ?

README.md
```diff
@@ -410,6 +410,39 @@ allow = response {
 }
 ```
 
+With denied response,
+array of headers can be passed to the downstream client.
```
srenatus, line 414: Can we add a short note on what the results are in the client?

Maybe something along the lines of
> With the following policy, a `POST /admin` request would result in the following response, featuring multiple `why` header values:
> ```
> 400 Bad Request
> why: POST method is not allowed
> why: /admin path is not allowed
>
> Unauthorized Request
> ```

internal/internal.go
```diff
@@ -18,6 +17,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/open-policy-agent/opa/topdown"
```
srenatus, line 20: I'd put it into the last lump to the other `github.com/open-policy-agent/opa` deps, but this is fine. ;)

internal/internal_test.go
srenatus: Well I guess we've got it covered by asserting that there's two headers. What else should they be than `bar` and `baz`? 🤔 

It's up to you, not blocking this from getting merged, but we could do something like (⚠️ untested sketch)
```go  
seen := map[string]int{}
for _, hdr := range result {
    seen[hdr.Header.Value]++
}
if seen["bar"] != 1 || seen["baz"] != 1 {
    t.Errorf("expected 'bar' and 'baz', got %v", seen)
}
```

srenatus (approved):
👍 This looks really good, thank you! I think one more tiny addition to the docs would help, though -- what do you think?

README.md
ashutosh-narkar: We should rephrase the text a little. Something like:

> The following policy shows how to generate an `object` response with the `headers` set as an array of values. The example policy restricts access to the `/admin` path and the `POST` method. 

<**add_ the_policy_example_here**>

The input to policy is show below:
```json
{
    	"attributes": {
        	"request": {
            	    "http": {
                	    "method": "POST",
                            "path": "/admin"
                    }
              }
        }
    }
```


The response generated by the policy for the above input looks like below:
```json
{
            "allowed": false,
            "body": "Unauthorized Request",
            "headers": [
                {
                    "why": "/admin path is not allowed"
                },
                {
                    "why": "POST method is not allowed"
                }
            ],
            "http_status": 400
}
```

ashutosh-narkar (approved):
LGTM. Thanks @keyolk !

PR #217, opened 2020-11-19
support ext_authz v3 API, still serve v2

srenatus:
This will make opa-envoy-plugin expose both the v2 and the v3 APIs:

```interactive
$ grpcurl -plaintext 127.0.0.1:9191 list
envoy.service.auth.v2.Authorization
envoy.service.auth.v3.Authorization
grpc.reflection.v1alpha.ServerReflection
```

As such, we should be able to use it with in either variant. New features will only be added to the v3 API, and v2 is to be considered deprecated, and will be removed in the future, timeline TBD.

There's still some work to do here:

- [x] figure out the impact of switching to [`protojson`](https://pkg.go.dev/google.golang.org/protobuf@v1.25.0/encoding/protojson) for converting the CheckRequest into OPA's input, and determine if we want to use v3 as a chance to do that.

  > This package produces a different output than the standard "encoding/json" package, which does not operate correctly on protocol buffer messages.
- [x] documentation: 
  - [x] what do I need to do to get to v3, (TODO: add istio)
  - [X] what changes in v3,
  - [X] did the check request change?
- [x] testing: the unit tests use v3, the e2e tests use v2 still

---

Fixes https://github.com/open-policy-agent/opa/issues/2798.


tsandall (commented):
Compatibility is definitely nice to have, so +1 there. I have two questions that I think we need to answer (which we've talked about offline):

* What config changes are required on the Envoy (and also Istio) side? E.g., does the filter configuration change significantly? An example would be useful.

* What policy changes are required when switching from v2 to v3? Does using protojson change anything? Examples would be useful.

internal/internal.go
```diff
@@ -223,25 +236,49 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx context.Context, req *ext_authz.Chec
...
+	var headers map[string]string
+
+	// NOTE: The path/body/headers blocks look silly, but they allow us to retrieve
+	//       the parts of the incoming request we care about, without having to convert
+	//       the entire v2 message into v3. It's nested, each level has a different type,
+	//       etc -- we only care for its JSON representation as fed into evaluation later.
+	switch r := r.(type) {
+	case *ext_authz_v3.CheckRequest:
+		req = r
+		bs, err = protojson.Marshal(req) // TODO(sr): Note that the encoding might have changed, figure out how.
```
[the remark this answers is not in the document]
srenatus: [The linked doc](https://developers.google.com/protocol-buffers/docs/proto3#json) explains how pb is supposed to be turned into json, but it doesn't show the differences from encoding/json... For example encoding/json won't match the spec'ed way to marshal a timestamp value:

```interactive
$ cat main.go
package main

import (
        "fmt"
        "time"
        "encoding/json"

        "google.golang.org/protobuf/types/known/timestamppb"
        "google.golang.org/protobuf/encoding/protojson"
)

func main() {
        ts := timestamppb.New(time.Now())

        bs0, err := json.Marshal(ts)
        if err != nil {
                panic(err)
        }

        bs1, err := protojson.Marshal(ts)
        if err != nil {
                panic(err)
        }

        fmt.Printf("encoding/json: %q\n", bs0)
        fmt.Printf("protojson: %q\n", bs1)
}
$ go run main.go
encoding/json: "{\"seconds\":1605811323,\"nanos\":912400000}"
protojson: "\"2020-11-19T18:42:03.912400Z\""
$
```

It's still logically possible that it doesn't matter how we marshal a CheckRequest (v2, v3), but it would be better to have a close look at all the values that can appear in attribute context. I'll take care of that as I compare v2 and v3 CheckRequests.

Either way, for v3, I think we should go with protojson, as it fulfils the protobuf specification. The question above is a) what this means concretely, and b) how big an impact it is for people writing policies against the JSON representation of a CheckRequest. (The intro of v3 is a good moment to fix this, since no user will switch accidentally from v2 to v3, but only on purpose.)

internal/internal.go
```diff
@@ -695,3 +727,54 @@ type internalError struct {
...
 	return e.Message
 }
+
+func (p *envoyExtAuthzV2Wrapper) Check(ctx context.Context, req *ext_authz_v2.CheckRequest) (*ext_authz_v2.CheckResponse, error) {
+	respV3, err := p.v3.check(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	return v2Response(respV3), nil
```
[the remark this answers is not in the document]
srenatus: I think it's negligible, but it would also be cleaner to include it in the measurements. 👍 Will do.

srenatus:
I've had a close look at the proto defs for v2 and v3, and played around some more, capturing my findings in https://github.com/open-policy-agent/opa-envoy-plugin/pull/217/files#diff-9119e7abecbea4bb21414d5dbc6de682261b4fc2840aec10485494feb01125dd

Summing up, I think it's preferable to go with protojson in v3; an alternative we could explore is adding a configurable to let people opt-in to the old behaviour, using encoding/json instead.

internal/internal.go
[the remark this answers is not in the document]
srenatus: I've dealt with that. It's not exactly beautiful but should be accurate. (And the little refactoring has let us drop the named returns in Check())

srenatus:
Remaining TODOs:

- [ ] figure out istio with v3, document it
- [ ] add e2e tests for v3
- [ ] determine if duplicating the existing unit tests (checking v3) for v2 is worth it

(💭 It's interesting how our current set of unit tests seems to live exactly in the overlap of v2/v3 _and_ encoding/json / protojson. While I'd like to expand them some day, I think it's better to focus on one set of those, and I'd think that's v3+protojson.)

README.md
```diff
@@ -461,6 +461,49 @@ allow {
 }
 ```
 
+## Envoy xDS v2 and v2
+
+This plugin exposes both versions. For v3 requests, the protojson encoding is used for making the
```
srenatus, line 466: todo: link to pb spec

srenatus:
✔️ ensured that `make test-e2e` still works. It turns out upgrading golang (#201) pulled in a change in `net/http` making the hardcoded TLS certs insufficient. CN in Subject no longer does the trick, we need a proper SNI record.

README.md
```diff
@@ -461,6 +461,49 @@ allow {
...
+incoming `envoy.service.auth.v3.CheckRequest` available in `input`. It differs from the encoding
+used for v2 requests:
+
+<!-- TODO(sr): table -->
```
tsandall, line 470: Once the table is done, I'd say this section looks great.

srenatus:
ℹ️ Rebasing now, I'll squash post-review.

srenatus:
Please note that I haven't updated the top-level `quick_start.yaml` file. It's using v2, and it should still be OK to do so.

README.md
```diff
@@ -519,14 +586,101 @@ been excluded for brevity):
...
-```ruby
+```rego
 jwt_payload = input.attributes.metadata_context.filter_metadata["envoy.filters.http.jwt_authn"].verified_jwt
 
 allow {
   jwt_payload.email == "alice@example.com"
 }
 ```
 
+## Envoy xDS v2 and v2
```
tsandall, line 597: This section is great.

internal/internal.go
```diff
@@ -718,3 +758,59 @@ type internalError struct {
...
+	respV3, stop, err := p.v3.check(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	respV2 := v2Response(respV3)
+	if code := stop(); code != nil {
+		respV2.Status = code
+	}
+	return respV2, nil
```
tsandall, line 773: I assume the conversion is relatively quick and accounts for a small portion (e.g., 1%) of the check response time...is that right? Since we're not including it in the decision log metrics, I just want to make sure it's negligible and can be ignored.

build/gen-tls-certs.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+cat > v3.txt <<- EOF
```
tsandall, line 3: I didn't see where this script was invoked from. Are we just using this to generate the cert that goes into the quick start? A comment explaining what this script is used for might be useful.

tsandall (approved):
Looks great. A couple questions but nothing blocking.

internal/internal.go
srenatus: It was my intention to _include it_ in the decision log time, that's where the awkward `stop` handling came from. It's converted in L769 right before stopping the time in L770. I might be mislead about how it works...? 😄

build/gen-tls-certs.sh
srenatus: 👍 Yup you're right. I find the generation of proper SNI TLS certs hard to remember (to say the least), and thus thought it better to record it like this. I'll add a comment!

srenatus:
⏸️ I'll address all comments and proceed with merging this tomorrow. 😃

internal/internal.go
tsandall: Whoops. I didn't look closely enough at what `stop()` was doing. The metric tracking looks fine then.

tsandall: However, we'll want to call `stop()` regardless of whether an error occurred--we want a decision log emitted in the case of errors.

srenatus: Yep, that's an oversight. We'll want `stop` to happen for the L756 case, too, indeed.

srenatus: I've fixed it, and duplicated a few of the unit tests for v2, to cover this logic.

srenatus:
I'll squash and merge. Let's take it from here in follow-up PRs.

PR #222, opened 2020-11-25
actions: fix e2e test, add it to pull-request workflow

srenatus:
Also updates kind to 0.9.0, required for Istio 1.7.0.
(Oversight of #217)

👉 Here's an example run of this PR: https://github.com/open-policy-agent/opa-envoy-plugin/runs/1453090996?check_suite_focus=true#step:4:211

⏱️ Just short of 5 minutes. Sounds OK to me. Let's try to roll with this in the pull-request workflow for a while.

### DIFF
--- a/.github/workflows/post-merge.yaml
+++ b/.github/workflows/post-merge.yaml
@@ -63,36 +63,26 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
 
-      - name: Run End-to-End Test
+      - name: Build docker image
+        run: make image tag-latest
+
+      - name: Setup kind/istio
         run: |
-          KIND_VERSION=0.4.0
-          ISTIO_VERSION=1.5.0
-
-          # Download and install kind
-          curl -L https://github.com/kubernetes-sigs/kind/releases/download/v${KIND_VERSION}/kind-linux-amd64 --output kind && chmod +x kind && sudo mv kind /usr/local/bin/
-
-          # Create kind cluster
-          if [ -z $(kind get clusters) ]; then
-            kind create cluster
-          fi
-          export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
-
-          # Download and install kubectl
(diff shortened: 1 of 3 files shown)


.github/workflows/post-merge.yaml
```diff
@@ -55,44 +55,3 @@ jobs:
         # Only run if required secrets are provided
         if: ${{ env.DOCKER_USER && env.DOCKER_PASSWORD }}
         run: make deploy-ci
```
ashutosh-narkar, line 57: Any reason for removing the e2e on master ? Even though we run it on a PR, it would fine to run it on master as well imo.

srenatus: Hmm good point. We can have both. I'll restore them.

build/install-istio-with-kind.sh
```diff
@@ -17,15 +17,16 @@ if [ -z $(kind get clusters) ]; then
...
 
+# Use kind cluster
+kubectl cluster-info --context kind-kind
+
 # Download and install Istio
 curl -L https://git.io/getLatestIstio | ISTIO_VERSION=${ISTIO_VERSION} sh - && mv istio-${ISTIO_VERSION} /tmp
-cd /tmp/istio-${ISTIO_VERSION}
+pushd /tmp/istio-${ISTIO_VERSION}
 bin/istioctl install -y --set profile=demo
+popd
```
srenatus, line 31: This `cd` to `pushd` change doesn't really matter for the script, but it also doesn't hurt...

.github/workflows/post-merge.yaml
srenatus: Updated, it's running twice now. Slightly refactored the actions in the process.

ashutosh-narkar (approved):
lgtm

srenatus:
✔️ looks OK here: https://github.com/open-policy-agent/opa-envoy-plugin/runs/1475406622

Thanks for the review 🚀

PR #223, opened 2020-11-30
workflow: use docker-env for building image (to ensure golang version)

srenatus:
Before, the `image` would invoke `build-linux`, and that wouldn't guarantee the
version of golang used for building.

Now, the `image` target goes through a round of `make release` to ensure just
that.

As a consequence, we'll build windows and darwin binaries where we don't need
them, but let's see if the time wasted is problematic at all.

To verify that this does what it's supposed to do, I've overridden the VERSION
makefile variable, and ran `make image`. The resulting container reports:

    $ docker run openpolicyagent/opa:"0.24.0-envoy-11" version
    Version: 0.24.0-envoy-11
    Build Commit: e4f19267
    Build Timestamp: 2020-11-30T10:40:36Z
    Build Hostname: 6bec0a11f7e7
    Go Version: go1.15.2

Fixes https://github.com/open-policy-agent/opa/issues/2942.

----
⚠️ There's one thing to note here -- actually running the service built using that Golang version means it'll be **less tolerant towards TLS certs than it was before**. In #217, I've run into this locally, so the `examples/istio/quick_start.yaml` certs have been adapted, but there hasn't been a release of opa-envoy-plugin using Go 1.15.x until we ship **this change**.

For more details about what's going on, [please refer to the release notes](https://golang.org/doc/go1.15#commonname):
> The deprecated, legacy behavior of treating the `CommonName` field on X.509 certificates as a host name when no `Subject Alternative Names` are present is now disabled by default. It can be temporarily re-enabled by adding the value `x509ignoreCN=0` to the `GODEBUG` environment variable.
>
> Note that if the `CommonName` is an invalid host name, it's always ignored, regardless of `GODEBUG `settings. Invalid names include those with any characters other than letters, digits, hyphens and underscores, and those with empty labels or trailing dots.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ VERSION_ISTIO := $(VERSION_OPA)-istio$(shell ./build/get-plugin-rev.sh)
 PACKAGES := $(shell go list ./.../ | grep -v 'vendor')
 
 GO := go
+GO_TAGS := -tags=opa_wasm
 GOVERSION := $(shell cat ./.go-version)
 GOARCH := $(shell go env GOARCH)
 GOOS := $(shell go env GOOS)
@@ -41,9 +42,10 @@ LDFLAGS := "-X github.com/open-policy-agent/opa/version.Version=$(VERSION) \
 GO15VENDOREXPERIMENT := 1
 export GO15VENDOREXPERIMENT
 
-.PHONY: all build build-darwin build-linux build-windows clean check check-fmt check-vet check-lint \
-    deploy-ci docker-login generate image image-quick push push-latest tag-latest \
-    test test-cluster test-e2e update-opa update-istio-quickstart-version version
+.PHONY: all build build-darwin build-linux build-windows clean check \
+    check-fmt check-vet check-lint deploy-ci docker-login generate image image-quick \
+    push push-latest tag-latest test test-cluster test-e2e update-opa \
+    update-istio-quickstart-version version
 
 ######################################################
 #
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -71,8 +71,8 @@ build-linux:
 build-windows:
 	@$(MAKE) build GOOS=windows
 
-image:
-	@$(MAKE) build-linux
+image: release
+	mv $(RELEASE_DIR)/opa_envoy_linux_$(GOARCH) .
```
tsandall, line 75: `make release` may produce a binary owned by root since it was built inside of Docker...perhaps we could just modify the Dockerfile to copy from the release directory instead of the root directory--that would avoid the `mv`.

tsandall (approved):
One minor comment about permissions on the binary.

RE: the cert/go version issue:  I think this is fine. I'm assuming this only applies to the certs in the quick_start.yaml file that are used for the admission controller (which is basically an installer for an example...)

I'm tempted to put a comment in the quick_start.yaml file that indices we don't guarantee backwards compatibility on the quick_start.yaml manifest (although that might get misinterpreted...)

Makefile
srenatus: I'll do that, the path I started on led to a mess, though. I'll clean it up tomorrow. We don't need multiple Dockerfiles, there's only one thing built here. 🤔

srenatus:
I'll fix e2e tomorrow, it'll be affected by #202 getting merged anyways, I suppose.

PR #227, opened 2020-12-15
internal: fix parsed path url-unescaping

srenatus:
Re-vived #175  to fix https://github.com/open-policy-agent/opa/issues/3004

Original PR message by @vetyy below:

------

Hello,

I found a problem with parsing path and query parameters.

The original test is actually invalid and is escaping reserved characters '=', '?' and '&' as per (https://tools.ietf.org/html/rfc3986#section-2.2).

If I change test to include a '\n' character in query parameter it will be first unescaped by url.PathUnescape and then again using url.Parse that will result in error "net/url: invalid control character in URL".

So there is no need to explicitly unescape path as url.Parse handles that internally.

I updated the code and test to be valid and I also tested in our own OPA deployment.

Thanks,
Matej

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -693,12 +693,7 @@ func getRevision(ctx context.Context, store storage.Store, txn storage.Transacti
 }
 
 func getParsedPathAndQuery(path string) ([]interface{}, map[string]interface{}, error) {
-	unescapedPath, err := url.PathUnescape(path)
-	if err != nil {
-		return nil, nil, err
-	}
-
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
LGTM

tsandall (commented):
LGTM. I took a quick look at the url.Parse implementation and it also calls unescape under the hood on the path. The different escape mode doesn't seem to have any affect in this case.

tsandall (approved):

ashutosh-narkar (approved):

PR #229, opened 2021-01-22
README, test-e2e: bump Istio to 1.8.x

srenatus:
👉  https://istio.io/latest/news/support/announcing-1.7-eol/

Istio 1.7.x goes EOL on Feb 19th. So, merging this isn't urgent, but if we want to bump, here it is.`test-e2e` run locally passed ✅

### DIFF
--- a/build/install-istio-with-kind.sh
+++ b/build/install-istio-with-kind.sh
@@ -7,7 +7,7 @@ set -x
 GOARCH=$(go env GOARCH)
 GOOS=$(go env GOOS)
 KIND_VERSION=0.9.0
-ISTIO_VERSION=1.7.0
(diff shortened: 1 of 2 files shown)


build/install-istio-with-kind.sh
```diff
@@ -7,7 +7,7 @@ set -x
 GOARCH=$(go env GOARCH)
 GOOS=$(go env GOOS)
 KIND_VERSION=0.9.0
-ISTIO_VERSION=1.7.0
+ISTIO_VERSION=1.8.2
```
srenatus, line 10: I think the latest minor version of the oldest supported series is a decent choice, what do you think?

ashutosh-narkar (approved):
lgtm

PR #230, opened 2021-02-10
internal: Include revisions of multiple bundles in decision log event

ashutosh-narkar:
Earlier the decision log events generated by the plugin did not set the
revision in cases where multiple bundles were configured. This change
includes these revisions in the log event and also adds a deprecation label
to the "revision" field of the eval result.

Fixes open-policy-agent/opa#3141

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -40,6 +40,7 @@ import (
 	"google.golang.org/protobuf/types/dynamicpb"
 
 	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/plugins"
 	"github.com/open-policy-agent/opa/plugins/logs"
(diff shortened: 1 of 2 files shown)


internal/internal.go
```diff
@@ -677,19 +684,31 @@ func uuid4() (string, error) {
...
-	if err != nil {
-		if storage.IsNotFound(err) {
-			return "", nil
+func getRevision(ctx context.Context, store storage.Store, txn storage.Transaction, result *evalResult) error {
+	revisions := map[string]string{}
+
+	names, err := bundle.ReadBundleNamesFromStore(ctx, store, txn)
+	if err != nil && !storage.IsNotFound(err) {
+		return err
+	}
```
srenatus, line 693: If `storage.IsNotFound(err)` is true, is it safe to use `names`? (It probably is, just wanting to be sure)

internal/internal.go
```diff
@@ -677,19 +684,31 @@ func uuid4() (string, error) {
...
+
+	for _, name := range names {
+		r, err := bundle.ReadBundleRevisionFromStore(ctx, store, txn, name)
+		if err != nil && !storage.IsNotFound(err) {
+			return err
 		}
```
srenatus, line 699: Same question as above, but for `r`

srenatus (approved):
LGTM! One question around the "not found" case, but it's not a blocker. 👍

internal/internal.go
ashutosh-narkar: I think it is.

PR #233, opened 2021-02-12
Extract reusable library

mjungsbluth:
In case you want to reuse the input and response handling structure outside of this plugin (i.e. with another reverse proxy), the current code structure doesn't allow re-use. 

This PR is meant a work in progress to discuss potential refactoring...

### DIFF
--- a/envoyauth/evaluation.go
+++ b/envoyauth/evaluation.go
@@ -1,0 +1,122 @@
+package envoyauth
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/metrics"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/storage"
+	iCache "github.com/open-policy-agent/opa/topdown/cache"
+	"github.com/sirupsen/logrus"
+)
+
+//EvalContext - This is an SPI that has to be provided if the envoy external authorization
+// is used from outside the plugin, i.e. as a Go module
+type EvalContext interface {
+	ParsedQuery() ast.Body
+	Store() storage.Store
+	Compiler() *ast.Compiler
+	Runtime() *ast.Term
+	PreparedQueryDoOnce() *sync.Once
+	InterQueryBuiltinCache() iCache.InterQueryCache
+	PreparedQuery() *rego.PreparedEvalQuery
+	SetPreparedQuery(*rego.PreparedEvalQuery)
+}
+
+//Eval - Evaluates an input against a provided EvalContext and yields result
+func Eval(ctx context.Context, evalContext EvalContext, input ast.Value, result *EvalResult, opts ...func(*rego.Rego)) error {
+
+	err := storage.Txn(ctx, evalContext.Store(), storage.TransactionParams{}, func(txn storage.Transaction) error {
+		err := getRevision(ctx, evalContext.Store(), txn, result)
+		if err != nil {
+			return err
+		}
+
+		result.TxnID = txn.ID()
+
+		logrus.WithFields(logrus.Fields{
+			"input": input,
+			"query": evalContext.ParsedQuery().String(),
+			"txn":   result.TxnID,
+		}).Debug("Executing policy query.")
+
+		err = constructPreparedQuery(evalContext, txn, result.Metrics, opts)
+		if err != nil {
+			return err
+		}
+
+		rs, err := evalContext.PreparedQuery().Eval(
+			ctx,
+			rego.EvalParsedInput(input),
+			rego.EvalTransaction(txn),
+			rego.EvalMetrics(result.Metrics),
+			rego.EvalInterQueryBuiltinCache(evalContext.InterQueryBuiltinCache()),
+		)
+
+		if err != nil {
+			return err
+		} else if len(rs) == 0 {
+			return fmt.Errorf("undefined decision")
+		} else if len(rs) > 1 {
+			return fmt.Errorf("multiple evaluation results")
+		}
+
+		result.Decision = rs[0].Expressions[0].Value
+		return nil
+	})
+
+	return err
+}
+
+func constructPreparedQuery(evalContext EvalContext, txn storage.Transaction, m metrics.Metrics, opts []func(*rego.Rego)) error {
+	var err error
+	var pq rego.PreparedEvalQuery
+
+	evalContext.PreparedQueryDoOnce().Do(func() {
+		opts = append(opts,
+			rego.Metrics(m),
+			rego.ParsedQuery(evalContext.ParsedQuery()),
+			rego.Compiler(evalContext.Compiler()),
+			rego.Store(evalContext.Store()),
+			rego.Transaction(txn),
+			rego.Runtime(evalContext.Runtime()))
+
+		r := rego.New(opts...)
+
+		pq, err = r.PrepareForEval(context.Background())
+		evalContext.SetPreparedQuery(&pq)
+	})
+
+	return err
+}
+
+func getRevision(ctx context.Context, store storage.Store, txn storage.Transaction, result *EvalResult) error {
+	revisions := map[string]string{}
+
+	names, err := bundle.ReadBundleNamesFromStore(ctx, store, txn)
+	if err != nil && !storage.IsNotFound(err) {
+		return err
+	}
+
+	for _, name := range names {
+		r, err := bundle.ReadBundleRevisionFromStore(ctx, store, txn, name)
+		if err != nil && !storage.IsNotFound(err) {
+			return err
+		}
+		revisions[name] = r
+	}
+
+	// Check legacy bundle manifest in the store
+	revision, err := bundle.LegacyReadRevisionFromStore(ctx, store, txn)
+	if err != nil && !storage.IsNotFound(err) {
+		return err
+	}
+
+	result.Revisions = revisions
+	result.Revision = revision
+	return nil
+}
--- a/envoyauth/evaluation_test.go
+++ b/envoyauth/evaluation_test.go
@@ -1,0 +1,104 @@
+package envoyauth
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/storage"
+	"github.com/open-policy-agent/opa/storage/inmem"
+)
+
+func TestGetRevisionLegacy(t *testing.T) {
+	store := inmem.New()
+	ctx := context.Background()
+
+	result := EvalResult{}
+
+	tb := bundle.Manifest{
+		Revision: "abc123",
+		Roots:    &[]string{"/a/b", "/a/c"},
+	}
+
+	// write a "legacy" manifest
+	err := storage.Txn(ctx, store, storage.WriteParams, func(txn storage.Transaction) error {
+		if err := bundle.LegacyWriteManifestToStore(ctx, store, txn, tb); err != nil {
+			t.Fatalf("Failed to write manifest to store: %s", err)
+			return err
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error finishing transaction: %s", err)
+	}
+
+	txn := storage.NewTransactionOrDie(ctx, store, storage.WriteParams)
+
+	err = getRevision(ctx, store, txn, &result)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expected := "abc123"
+	if result.Revision != "abc123" {
+		t.Fatalf("Expected revision %v but got %v", expected, result.Revision)
+	}
+
+	if len(result.Revisions) != 0 {
+		t.Fatal("Unexpected multiple bundles")
+	}
+}
+
+func TestGetRevisionMulti(t *testing.T) {
+	store := inmem.New()
+	ctx := context.Background()
+
+	result := EvalResult{}
+
+	bundles := map[string]bundle.Manifest{
+		"bundle1": {
+			Revision: "abc123",
+			Roots:    &[]string{"/a/b", "/a/c"},
+		},
+		"bundle2": {
+			Revision: "def123",
+			Roots:    &[]string{"/x/y", "/z"},
+		},
+	}
+
+	// write bundles
+	for name, manifest := range bundles {
+		err := storage.Txn(ctx, store, storage.WriteParams, func(txn storage.Transaction) error {
+			err := bundle.WriteManifestToStore(ctx, store, txn, name, manifest)
+			if err != nil {
+				t.Fatalf("Failed to write manifest to store: %s", err)
+			}
+			return err
+		})
+		if err != nil {
+			t.Fatalf("Unexpected error finishing transaction: %s", err)
+		}
+	}
+
+	txn := storage.NewTransactionOrDie(ctx, store, storage.WriteParams)
+
+	err := getRevision(ctx, store, txn, &result)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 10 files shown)


anderseknert:
I've just started looking into https://github.com/open-policy-agent/opa/issues/3140 and it looks like much of `evaluation.go` could be covered in OPA directly. Perhaps we can sync on that once I have some code to show :)

mjungsbluth:
Yes, that was also my feeling. Mind that most of my changes have been mechanical and moved existing code in this plugin around. 
To see what changed I‘d recommend to go commit by commit(every step builds successfully).

anderseknert:
@mjungsbluth see https://github.com/open-policy-agent/opa/pull/3186

mjungsbluth:
@anderseknert had a look at the referenced PR which looks promising to simplify this plugin. 
Regarding this PR: I don‘t think we would need to wait on changes in OPA to merge this. The changes in this PR probably even make it simpler to delete portions of code as soon as they come as part of opa itself. My experience with PRs this size is just that waiting normally leads to a lot of work to keep them updated. So some feedback if this is roughly ok would be appreciated (I would then try to integrate the changes and see if I can remove all/most code duplication).

Also a bit more on the use case: We use Skipper instead of Envoy as both Ingress and for general purpose reverse proxing. Skipper is itself build in Golang so we can both optimize latency and memory consumption by sharing bundles across different target applications. Evaluation would be executed as a Skipper filter. We are also a Styra customer so would like to reuse all structures like Stacks that have already been built for envoy which effectively means we want to use the same input + decision structures. I integrated everything in a PoC by copying large portions of this plugin that were private or too tied to grpc.

anderseknert:
Interesting use case @mjungsbluth! And I agree - non need to wait for changes in OPA for that. We can always refine this later once those land. I'll review tomorrow.

internal/envoyauth/evaluation.go
```diff
@@ -0,0 +1,147 @@
...
+//EvalResult - Captures the result from evaluating a query against an input
+type EvalResult struct {
+	Revision   string // Deprecated: Use `revisions` instead.
+	Revisions  map[string]string
+	DecisionID string
+	TxnID      uint64
+	Decision   interface{}
+	Metrics    metrics.Metrics
+}
+
```
ashutosh-narkar, line 39: Nit: This can be moved to `response.go`

internal/envoyauth/response.go
```diff
@@ -0,0 +1,165 @@
...
+
+// IsAllowed - Returns if the decision is representing an "allow" depending on the decision structure.
+// Returns an error if the decision structure is invalid
+func (result *EvalResult) IsAllowed() (bool, error) {
+	err := result.IsDecisionValid()
+
+	if err != nil {
+		return false, err
+	}
+
```
ashutosh-narkar, line 20: Nit: Adding a switch case here should be fine. Also `IsDecisionValid()` can be moved in this module and can be private.

ashutosh-narkar (commented):
This re-org looks pretty good ! I wonder if there are some common methods that can be moved into the `util` package. @mjungsbluth have you tried to integrate with this refactored version of the plugin ?

mjungsbluth:
@ashutosh-narkar there are some methods like the uuid4 method that can go to utils as well I think. Waited for a general ok on this PR before integrating to prevent double work. Will start with it now and update this PR afterwards (also adressing your comments)

internal/envoyauth/evaluation.go
mjungsbluth: Done.

mjungsbluth:
@ashutosh-narkar I did reintegrate and while that itself went smoothly, there were some changes that I had to do:
* Move packages out of internal
* The http status calculation was still envoy grpc specific
* The GetResponse* methods exposed the internal decision structure, I changed them so this knowledge now stays internal

internal/internal.go
```diff
@@ -244,21 +239,18 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 	var err error
 	start := time.Now()
 
-	result := evalResult{}
-	result.metrics = metrics.New()
-	result.metrics.Timer(metrics.ServerHandler).Start()
-	result.decisionID, err = uuid4()
+	result, stopeval, err := envoyauth.NewEvalResult()
 	if err != nil {
 		logrus.WithField("err", err).Error("Unable to generate decision ID.")
```
ashutosh-narkar, line 244: This error message should be updated.

internal/internal.go
```diff
@@ -273,88 +265,44 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
...
-		status, err := getResponseStatus(decision)
-		if err != nil {
-			return nil, stop, errors.Wrap(err, "failed to get response status")
-		}
+	if err != nil {
+		return nil, stop, errors.Wrap(err, "failed to get response status")
+	}
 
-		resp.Status = &rpc_status.Status{Code: status}
+	status := int32(code.Code_PERMISSION_DENIED)
```
ashutosh-narkar, line 297: Nit: We could add a switch case here.

ashutosh-narkar (commented):
@mjungsbluth this is looking good. Left a few comments.

envoyauth/response.go
```diff
@@ -0,0 +1,292 @@
...
+	DecisionID string
+	TxnID      uint64
+	Decision   interface{}
+	Metrics    metrics.Metrics
+}
+
+//StopFunc should be called as soon as the evaluation is finished
+type StopFunc = func()
+
+//NewEvalResult creates a new EvalResult and and a StopFunc that is used to stop the timer for metrics
```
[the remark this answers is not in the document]
mjungsbluth: That is really your call. It would essentially mean that users of this method must know that they have to start + stop the timer on the metrics object so that the results show up in the decision log to my understanding. It was code I had to copy in my use of this codebase. The way it is now, I was forced to think about calling the stop function.

envoyauth/response.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		if allowed, ok = val.(bool); !ok {
+			return false, fmt.Errorf("type assertion error")
+		}
+
+		return allowed, nil
+	}
+
+	return false, fmt.Errorf("illegal value for policy evaluation result: %T", result.Decision)
```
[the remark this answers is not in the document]
mjungsbluth: In short the golang compiler forces you to handle this cases (that should never happen). That is the reason I didn't have a switch before as the code was slightly longer (3 lines) but more readable.

envoyauth/response.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		return responseHeaders, nil
+	}
+
+	return nil, fmt.Errorf("illegal value for policy evaluation result: %T", result.Decision)
+}
+
+// GetResponseEnvoyHeaderValueOptions - returns the http headers to return if they are part of the decision as envoy header value options
+func (result *EvalResult) GetResponseEnvoyHeaderValueOptions() ([]*ext_core_v3.HeaderValueOption, error) {
```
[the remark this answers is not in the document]
mjungsbluth: One method returns `http.Headers` which is for example we use in Skipper and is not Envoy specific. The other method is for use in this plugin.

envoyauth/response.go
```diff
@@ -0,0 +1,292 @@
...
+
+		if http.StatusText(int(httpStatusCode)) == "" {
+			return status, fmt.Errorf("Invalid HTTP status code %v", httpStatusCode)
+		}
+
+		return int(httpStatusCode), nil
+	}
+
+	return http.StatusForbidden, fmt.Errorf("illegal value for policy evaluation result: %T", result.Decision)
+}
```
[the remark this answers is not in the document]
mjungsbluth: Sure, it's just that when you ever want to extend the decision structures it is much easier to enforce this in one place (IsDecisionValid) than changing all places where the decision is restructures (this is how it was before in this code base). Also error handling is now guaranteed resulting in the same message. Again, your call...

ashutosh-narkar:
@mjungsbluth let us know if we can make a pass over the changes or if you have some pending ?

envoyauth/response.go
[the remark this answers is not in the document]
mjungsbluth: After our discussion removed the `IsDecisionValid`

envoyauth/response.go
```diff
@@ -0,0 +1,292 @@
...
+			}
+
+		default:
+			return nil, fmt.Errorf("type assertion error")
+		}
+
+		return responseHeaders, nil
+	}
+
+	return nil, fmt.Errorf("illegal value for policy evaluation result: %T", result.Decision)
```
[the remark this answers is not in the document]
mjungsbluth: After our discussion removed the `IsDecisionValid`

envoyauth/response.go
```diff
@@ -0,0 +1,292 @@
...
+			responseHeaders = append(responseHeaders, headerValueOption)
+		}
+	}
+
+	return responseHeaders, nil
+}
+
+// HasResponseBody returns true if the decision defines a body (only true for structured decisions)
+func (result *EvalResult) HasResponseBody() (bool, error) {
+	err := result.IsDecisionValid()
```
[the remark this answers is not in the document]
mjungsbluth: After our discussion removed the `IsDecisionValid`

envoyauth/response.go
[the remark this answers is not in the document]
mjungsbluth: After our discussion removed the `IsDecisionValid`

internal/internal.go
mjungsbluth: Done.

mjungsbluth:
> @mjungsbluth let us know if we can make a pass over the changes or if you have some pending ?

@ashutosh-narkar you can give it a final pass, I addressed the comments

mjungsbluth:
I will also sign-off the commits once there are no further comments to address...

envoyauth/evaluation.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+//Eval - Evaluates an input against a provided EvalContext and yields result
+func Eval(ctx context.Context, evalContext EvalContext, input ast.Value, result *EvalResult, opts ...func(*rego.Rego)) error {
+
+	err := storage.Txn(ctx, evalContext.Store(), storage.TransactionParams{}, func(txn storage.Transaction) error {
+
+		var err error
+
+		err = getRevision(ctx, evalContext.Store(), txn, result)
```
srenatus, line 37: [nit]
```suggestion
		err := getRevision(ctx, evalContext.Store(), txn, result)
```

envoyauth/evaluation.go
```diff
@@ -0,0 +1,125 @@
...
+			return err
+		}
+
+		result.TxnID = txn.ID()
+
+		logrus.WithFields(logrus.Fields{
+			"input": input,
+			"query": evalContext.ParsedQuery().String(),
+			"txn":   result.TxnID,
+		}).Debug("Executing policy query.")
```
srenatus, line 48: 💭 I think we should go with "lowercase first letter, no period at the end" eventually, but not this time.

envoyauth/evaluation.go
```diff
@@ -0,0 +1,125 @@
...
+
+		err = getRevision(ctx, evalContext.Store(), txn, result)
+		if err != nil {
+			return err
+		}
+
+		result.TxnID = txn.ID()
+
+		logrus.WithFields(logrus.Fields{
+			"input": input,
```
srenatus, line 45: 💭 `input` could be pretty huge. But it's debug, so, fine.

envoyauth/evaluation_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	txn := storage.NewTransactionOrDie(ctx, store, storage.WriteParams)
+
+	err := getRevision(ctx, store, txn, &result)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(result.Revisions) != 2 {
+		t.Fatalf("Expected two bundles but got %v\n", len(result.Revisions))
```
srenatus, line 92: [nit] There's no need to add newlines to `t.Fatal` and the like, they'll have one appended if not present.

envoyauth/response.go
```diff
@@ -0,0 +1,254 @@
...
+type EvalResult struct {
+	Revision   string // Deprecated: Use `revisions` instead.
+	Revisions  map[string]string
+	DecisionID string
+	TxnID      uint64
+	Decision   interface{}
+	Metrics    metrics.Metrics
+}
+
+//StopFunc should be called as soon as the evaluation is finished
```
srenatus, line 24: [nit] I'd probably add a whitespace after every `//`. But it's just another these silly little code-style comments, feel free to ignore.

envoyauth/response_test.go
```diff
@@ -0,0 +1,186 @@
...
+	input["allowed"] = true
+	var result bool
+	result, err = er.IsAllowed()
+
+	if err != nil {
+		t.Fatalf("Expected no error but got %v", err)
+	}
+
+	if result != true {
+		t.Fatalf("Expected result %v but got %v", int32(code.Code_OK), result)
```
srenatus, line 38: Does this error message match what we test for? It looks like we're checking if `result` is true or not, and not whether it is `int32(code.Code_OK)`. 🤔

srenatus (approved):
👏 Nice refactoring. I've got a few minor comments, feel free to ignore the code-style related ones. The error-checking related question (please see inline) might be worth discussing.

That aside, LGTM.

envoyauth/evaluation.go
```diff
@@ -0,0 +1,125 @@
...
+			rego.ParsedQuery(evalContext.ParsedQuery()),
+			rego.Compiler(evalContext.Compiler()),
+			rego.Store(evalContext.Store()),
+			rego.Transaction(txn),
+			rego.Runtime(evalContext.Runtime()))
+
+		r := rego.New(opts...)
+
+		pq, err = r.PrepareForEval(context.Background())
+		evalContext.SetPreparedQuery(&pq)
```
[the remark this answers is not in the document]
mjungsbluth: I was very careful not to change behavior, so this actually already exists in master: https://github.com/open-policy-agent/opa-envoy-plugin/blob/42b3ce43573c94b02259a2b020bcdae3cb24c108/internal/internal.go#L482

I probably have even less context so this is your call.

envoyauth/evaluation.go
mjungsbluth: Also not touched by this PR: https://github.com/open-policy-agent/opa-envoy-plugin/blob/42b3ce43573c94b02259a2b020bcdae3cb24c108/internal/internal.go#L437

envoyauth/evaluation.go
mjungsbluth: Same here: https://github.com/open-policy-agent/opa-envoy-plugin/blob/42b3ce43573c94b02259a2b020bcdae3cb24c108/internal/internal.go#L432-L433

envoyauth/evaluation_test.go
mjungsbluth: Changed. But again, not change to existing code base

envoyauth/response.go
mjungsbluth: Changed

envoyauth/response_test.go
mjungsbluth: Good catch!

envoyauth/evaluation.go
mjungsbluth: Changed

srenatus:
Thanks for bearing with me. Let's rebase and get this in quick 🚀

mjungsbluth:
@srenatus rebased cleanly on the current master

srenatus:
@mjungsbluth Sorry to bother again, but would you mind squashing your commits, too? 😅 (We've got squash and merge disabled in this repo -- alternatively, I'll wait for someone to change the setting...)

mjungsbluth:
@srenatus done.

PR #234, opened 2021-02-13
Gloo Edge API Gateway integration example and tutorial.

eshepelyuk:
Inspired by this [Slack converstion](https://openpolicyagent.slack.com/archives/C7Y04PF50/p1612874017016900?thread_ts=1612864329.015800&cid=C7Y04PF50)

### DIFF
--- a/examples/gloo-edge/README.md
+++ b/examples/gloo-edge/README.md
@@ -1,0 +1,145 @@
+# Using OPA with Gloo Edge
+
+* [Gloo Edge](https://docs.solo.io/gloo-edge/latest/)
+* [OPA](https://www.openpolicyagent.org/docs/latest/envoy-authorization/)
+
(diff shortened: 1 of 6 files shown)


eshepelyuk:
@srenatus please check

ashutosh-narkar (commented):
The content looks great @eshepelyuk ! It would be helpful if we could re-org the headings based on some of the other examples we have. Something like:

* Overview
* How does it work ?
* Steps
   1. Start Minikube
   2. Similar to the steps in [this](https://www.openpolicyagent.org/docs/latest/envoy-authorization/#steps) tutorial.

eshepelyuk:
Hello @ashutosh-narkar 

Firts of all - I dont use minikube, and this tutorial doesn't depend on minikube and could be run at any k8s cluster.

Then, pleae feel free to arrange the text as you consider if would look better. I think this will be a most productive way of proceeding with this PR.

ashutosh-narkar (approved):

PR #238, opened 2021-03-04
ci: use vendor/ only, bump go

srenatus:
Also makes the go version be read from `go.mod`, a neat trick shared with me by @johanneslarsson.


tsandall (approved):
LGTM

PR #247, opened 2021-05-23
Update opa-istio example to allow control of the injection policy at the pod level.

MugiOka:
Did you not need to update this example?

I thought it would be more helpful to developers if it was explicitly stated in the example, because in my environment, I need to control the injection policy at the pod level.

### DIFF
--- a/examples/istio/README.md
+++ b/examples/istio/README.md
@@ -90,6 +90,18 @@ IP/port of the Istio Ingress gateway.
     curl --user bob:password -i http://$GATEWAY_URL/productpage
     curl --user bob:password -i http://$GATEWAY_URL/api/v1/products
     ```
+## Controlling the injection policy at the pod level
+
+If you want to control the injection policy at the pod level, set the `sidecar.opa-istio.io/inject` label to `false` on the pod.
+An example of the updated admission controller configuration is shown below:
+```yaml
+objectSelector:
+  matchExpressions:
+  - key: sidecar.opa-istio.io/inject
+    operator: NotIn
+    values:
+    - "false"
+```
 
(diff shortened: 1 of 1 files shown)


ashutosh-narkar:
Thanks for the contribution @MugiOka ! Let's add a note in the Readme like you've done and follow that with the updated admission controller config. For example:

```bash
If you want to control the injection policy at the pod level, set the sidecar.opa-istio.io/inject label to false on the pod. An example of the updated admission controller configuration is shown below:
```

```yaml
objectSelector:
      matchExpressions:
      - key: sidecar.opa-istio.io/inject
        operator: NotIn
        values:
        - "false"
```

examples/istio/quick_start.yaml
```diff
@@ -364,6 +364,12 @@ webhooks:
     namespaceSelector:
       matchLabels:
         opa-istio-injection: enabled
+    objectSelector:
+      matchExpressions:
+      - key: sidecar.opa-istio.io/inject
+        operator: NotIn
+        values:
+        - "false"
     failurePolicy: Fail
```
ashutosh-narkar, line 373: Let's revert the changes in the quick start as we've already documented this in the README.

ashutosh-narkar:
@MugiOka just one small change. Then please squash your commits and we can merge ! Thanks !

examples/istio/quick_start.yaml
MugiOka: Oh, I'm sorry.
I fixed it.

ashutosh-narkar (approved):
LGTM ! Thanks @MugiOka !

PR #253, opened 2021-06-17
add log messages for unsupported/missing content-type

justincely:
Signed-off-by: Justin Ely <justincely@gmail.com>

Adds log messages to help users debug if they aren't seeing the `parsed_body` field populated correctly.  

Closes https://github.com/open-policy-agent/opa/issues/3566

### DIFF
--- a/envoyauth/request.go
+++ b/envoyauth/request.go
@@ -160,7 +160,11 @@ func getParsedBody(logEntry *logrus.Entry, headers map[string]string, body strin
 			if !known {
 				return nil, false, nil
 			}
+		} else {
+			logEntry.Debugf("content-type: %s parsing not supported", val)
 		}
+	} else {
+		logEntry.Debug("no content-type header supplied, performing no body parsing")
 	}
 
 	return data, false, nil


envoyauth/request.go
```diff
@@ -160,7 +160,11 @@ func getParsedBody(logEntry *logrus.Entry, headers map[string]string, body strin
 			if !known {
 				return nil, false, nil
 			}
+		} else {
+			logEntry.Infof("content-type: %s parsing not supported", val)
```
ashutosh-narkar, line 164: What about keeping this at debug level ?

justincely: yeah, i wasn't quite sure where to put this when i did it. On one hand someone supplying a common content type and not having it parsed may warrant something at Info, but for a more obscure type Debug seems fine.  I went with the verbose option to maybe stave off user-confusion.

That being said, my opinion on that isn't strong; i'll happily drop it to debug if you think that's better.

ashutosh-narkar:
Thanks for making the changes @justincely. Can you please squash your commits and we can merge.

justincely:
sure thing @ashutosh-narkar, done!

PR #255, opened 2021-06-22
deps: bump OPA to v0.29.4

srenatus:
Something seemed to have gone wrong in the auto-updater. Trying to do
this manually.


ashutosh-narkar:
Are we trying out dependabot  as per the changes [here](https://github.com/open-policy-agent/opa-envoy-plugin/pull/257) first ?

PR #256, opened 2021-06-22
Fix unmarshal json when Content-Type is application/json

Dakatan:
It should not be able to parse yaml when Content-Type is application/json.
However `util.Unmarshal` is a utility that can parse yaml or json,
so if you put yaml in the body when Content-Type is application/json, it will be parsed.
Therefore, change the calling function to a `util.UnmarshalJSON`, and make it possible to parse only json.

In addition, `util.Unmarshal` was not possible to unmarshal a json string containing `\/` (which is an escaped `/`).
Json is supposed to consider slashes (`/`) as escapable characters (not required).
It could not parse a json body containing string of `\/`.
Because of the above, the opa-envoy-plugin eventually returned 403 whenever the Content-Type was application/json and the json contained `\/` in the string.
You can also avoid this problem by using the `util.UnmarshalJSON`.

Reference information
https://github.com/open-policy-agent/opa/pull/3571

### DIFF
--- a/envoyauth/request_test.go
+++ b/envoyauth/request_test.go
@@ -73,7 +73,7 @@ func TestGetParsedBody(t *testing.T) {
 			  "headers": {
 				"content-type": "application/json"
 			  },
-			  "body": "foo"
+			  "body": "\"foo\""
 			}
 		  }
 		}
@@ -171,12 +171,26 @@ func TestGetParsedBody(t *testing.T) {
 		}
 	  }`
 
+	requestBodyWithJSONSpecialChars := `{
+		"attributes": {
+		  "request": {
+			"http": {
+			  "headers": {
+				"content-type": "application/json"
+			  },
+			  "body": "[\"\\\"\", \"\\\\\", \"\\/\", \"/\", \"\\b\", \"\\f\", \"\\n\", \"\\r\", \"\\t\", \"\\u0041\"]"
+			}
+		  }
+		}
+	  }`
+
 	expectedNumber := json.Number("42")
 	expectedObject := map[string]interface{}{
 		"firstname": "foo",
 		"lastname":  "bar",
 	}
 	expectedArray := []interface{}{"hello", "opa"}
+	expectedJSONSpecialChars := []interface{}{`"`, `\`, "/", "/", "\b", "\f", "\n", "\r", "\t", "A"}
 
 	tests := map[string]struct {
 		input           *ext_authz.CheckRequest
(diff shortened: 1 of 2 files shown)


envoyauth/request_test.go
```diff
@@ -73,7 +73,7 @@ func TestGetParsedBody(t *testing.T) {
 			  "headers": {
 				"content-type": "application/json"
 			  },
-			  "body": "foo"
+			  "body": "\"foo\""
```
srenatus, line 76: 💭 this is interesting... I think it's correct _this way_, but would this mean that the behaviour changes in some what that's unrelated to escaped characters...?

srenatus (commented):
Thanks a lot for your work here. Tests included as well, 🎉 👏 

One question inline, and if you wouldn't mind, could you mention the forward-slash-escaping problem you've had originally in the commit message? Right now, it sounds like we'd only want to forbid yaml, which isn't entirely true. 😄

envoyauth/request_test.go
Dakatan: Yes, that's true, but if foo is directly in the body, I don't think it's a json literal (it looks like `true` or `false`).
The reason why it was working so far is that util.Unmarshal assumed it was yaml and converted `foo` to `"foo"`. (It seems that yaml considers `foo` as a string even without quoting).
Thus, I thought that `foo` would not pass the test, but should be converted to `"foo"`.

srenatus: Ah, gotcha. So, having a plain "foo" body (not "\"foo\"") actually is a case of sending YAML (not JSON).  And that's unlikely to be sent by anyone using an application/json header anyways.

I think we're OK here, we should, however, point out that we had erroneously been accepting YAML with a JSON header in the next _Release Notes_ 🎗️

Dakatan:
@srenatus 
Is the `forward-slash-escaping problem` you are talking about the following...?

> In addition, since the util.Unmarshal does not consider / as an escapable character (escaping / is optional in the json specification),
it could not parse json with \/ in the string, there was also the problem of getting all result 403.

If so, I think you can write in some detail.

envoyauth/request_test.go
Dakatan: You're right.
The release notes seem to use the github feature.
Is there anything I can do to help?

srenatus: It's cool, thanks, I'll take care of making a note for it. 😃

Dakatan:
@srenatus 
I've updated it to the commit message and PR content.
What do you think of the content?
I think it's a bit long... 🤔

srenatus (approved):
Thanks again!

ashutosh-narkar:
@srenatus can you add a `CHANGELOG.md` to keep track of this change ? We can later add it to the release notes as well.

srenatus:
@ashutosh-narkar here we go: https://github.com/open-policy-agent/opa-envoy-plugin/pull/260

PR #257, opened 2021-06-22
deps: enable dependabot

srenatus:


srenatus:
👉 Note: we've got to adapt our "replace all version strings" logic, which is something the service updater did, but that dependabot can't do.

srenatus:
I can take care of the version-string-sed'ing in a follow-up PR -- enabling dependabot shouldn't have any negative effects right now, it still depends on us merging its PRs. So, there should be no harm in merging this as-is now.

ashutosh-narkar:
Good to know !
> enabling dependabot shouldn't have any negative effects right now, it still depends on us merging its PRs. So, there should be no harm in merging this as-is now.

ashutosh-narkar (approved):

PR #259, opened 2021-06-22
build: update tag script

srenatus:
This allows us to use dependabot for updating OPA.

For any merge, we'll check if the version of OPA has changed,
and if so, push a commit fixing the version tags in the example
yamls.

It's done in a rather blunt manner: we replace whatever version
is in the yamls with whatever version we have for opa in go.mod,
and if the result is different from what we had before, we'll
push it to main.

### DIFF
--- a/.github/workflows/post-merge.yaml
+++ b/.github/workflows/post-merge.yaml
@@ -1,0 +1,41 @@
+name: Post Merge
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  generate:
+    name: Update versions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Commit & Push
+        shell: bash
+        run: |
+          # Commit any changes and push as needed.
+
+          # See https://github.com/actions/checkout#push-a-commit-using-the-built-in-token
+          AUTHOR=version-tag-updater
+          git config user.name ${AUTHOR}
+          git config user.email ${AUTHOR}@github.com
+
+          # Prevent looping if the build was non-deterministic..
+          CAN_PUSH=1
+          if [[ "$(git log -1 --pretty=format:'%an')" == "${AUTHOR}" ]]; then
+              CAN_PUSH=0
+          fi
+
+          if ./build/update-opa-version.sh; then
+            if [[ "${CAN_PUSH}" == "1" ]]; then
+              git push
+            else
+              echo "Previous commit was auto-generated -- Aborting!"
+              exit 1
+            fi
+          else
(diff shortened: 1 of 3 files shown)


.github/workflows/post-merge.yaml
```diff
@@ -0,0 +1,41 @@
...
+  generate:
+    name: Update versions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Commit & Push
+        shell: bash
+        run: |
```
srenatus, line 18: This was taken from OPA, https://github.com/open-policy-agent/opa/blob/main/.github/workflows/post-merge.yaml#L21-L46

build/update-opa-version.sh
```diff
@@ -1,43 +1,36 @@
...
-  sed -i.bak "/opa_container/{N;s/openpolicyagent\/opa:.*/openpolicyagent\/opa:$tag-istio\"\,/;}" examples/istio/quick_start.yaml && rm examples/istio/quick_start.yaml.bak
-  sed -i.bak "s/image: openpolicyagent\/opa:.*/image: openpolicyagent\/opa:$tag/" examples/istio/quick_start.yaml && rm examples/istio/quick_start.yaml.bak
+if [[ -z "$(git diff --name-only --cached)" ]]; then
+  echo "No version changes to commit!"
+  exit 1
+fi
 
-  # update vendor
-  env GO111MODULE=on go mod vendor
+git commit -m "examples: update version tags ($tag)"
```
ashutosh-narkar, line 31: Nit: `update OPA version tags`

Makefile
```diff
@@ -106,9 +106,6 @@ docker-login:
 
 deploy-ci: docker-login image push tag-latest push-latest
 
-update-opa:
```
ashutosh-narkar, line 109: This target is currently used by the auto-revendor service. Do we plan to update that ?

ashutosh-narkar (commented):
Changes look good. Few nits. Also do we have an example test run (for ex. on your fork) which shows how this will work and we get the expected results.

Makefile
srenatus: That revendor service would be removed when we switch to using this script and dependabot.

srenatus:
@ashutosh-narkar I've run it on my fork, the results were shared in slack. 😃

Makefile
ashutosh-narkar: Great !

ashutosh-narkar:
> @ashutosh-narkar I've run it on my fork, the results were shared in slack. 😃

Awesome !

ashutosh-narkar (approved):
LGTM

PR #262, opened 2021-06-26
build: Update commit author in plugin rev script

ashutosh-narkar:
Earlier opa-updater-automation was the author for
updating the OPA version in the example yamls and README.
After we changed to use dependabot for updating OPA, the examples
are now updated by the version-tag-updater author.


srenatus:
Thank you!

PR #263, opened 2021-06-27
Update docker image tag

shiv3:
When I tried to do a quick start, I couldn't pull the docker image.
It didn't seem to be pushed [here](https://hub.docker.com/r/openpolicyagent/opa/tags?page=1&ordering=last_updated&name=0.29.4-envoy), but isn't this version? : `0.29.4-envoy-2`

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ To deploy OPA-Envoy include the following container in your Kubernetes Deploymen
 
 ```yaml
 containers:
-- image: openpolicyagent/opa:0.29.4-envoy
+- image: openpolicyagent/opa:0.29.4-envoy-2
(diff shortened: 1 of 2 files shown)


srenatus (approved):
Yup, I believe this was a side-effect of having forgotten #262 when doing #259. Thanks!

srenatus:
I believe the same is true for the `istio` tag. Would you mind fixing the references to `opa:0.29.4-istio`, too? 😃

srenatus:
It's cool, I'll deal with istio. Thanks for raising and fixing this! 🚀

srenatus:
👉 #264

PR #264, opened 2021-06-29
workflow/checks: run istio example with tagged image

srenatus:
🚧 ~I want to see this fail first, I'll fix the tag next.~

✔️ Test fails on main, passes on this PR. Timeout set to 10min which be enough to only ever be triggered by problems, never by accident.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -61,7 +61,13 @@ jobs:
         run: make deploy-ci
 
   e2e:
-    name: Istio End-to-End Test
+    name: Istio End-to-End Test (${{ matrix.variant }})
+    strategy:
+      matrix:
+        include: # run tests with opa-envoy-plugin docker image built from...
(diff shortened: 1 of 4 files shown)


srenatus:
Oh my that's annoying. It'll fail but it takes loooong.

.github/workflows/checks.yaml
```diff
@@ -61,7 +61,13 @@ jobs:
         run: make deploy-ci
 
   e2e:
-    name: Istio End-to-End Test
+    name: Istio End-to-End Test (${{ matrix.variant }})
+    strategy:
+      matrix:
+        include:
+          - variant: main
+          - variant: tag
```
ashutosh-narkar, line 69: Can you add a comment describing what this means ?

ashutosh-narkar (commented):
Changes look good to me.

.github/workflows/checks.yaml
srenatus: Will do!

PR #267, opened 2021-07-04
deps: bump OPA

srenatus:


srenatus:
Hehe. There's trouble with the recently introduced other check...which cannot work because there's no tag yet... I'll disable it for now, I think...

srenatus:
So, on revisiting the tagging/release process here. @ashutosh-narkar please correct me if I'm wrong, but right now:

1. When bumping OPA via dependabot (like before service-updater), the result of `build/get-plugin-rev.sh` changes: it counts the number of commits since there was a dependabot (previously service-updater) merge.
2. That `-#` prefix, plus the OPA version as reported by `build/get-opa-version.sh`, make up the tag used for the docker images built, and inserted into the binaries.
3. On **merge**, the docker images a pushed to dockerhub.
4. On **tag-push**, there's a github release getting created, from **the pushed tag**.

Looks like the pushed tag could differ from the constructed tag that's used with dockerhub...? It should never do that, but it could.

Some questions here:

❓ I wonder if we cannot change this process slightly, so that tagged docker images are only ever pushed for the `tag-push` workflow, i.e., when we also create a github release?
❔ Do we need to publish binaries at all? Do we know if anyone is using them? I'd suspect not, at the very least, the darwin and windows binaries seem superfluous.

I think I'd prefer to fully go either way on the first issue: either let's tag and release `OPA_VERSION-envoy-REV` automatically for each merge, or don't do anything on merge and push releases to github and dockerhub based on the _chosen_ tag pushed when we decide to release opa-envoy-plugin. What do you think?

ashutosh-narkar:
> either let's tag and release OPA_VERSION-envoy-REV automatically for each merge

We already create the `OPA_VERSION-envoy-REV` tag after each merge. If we automatically create releases after each merge it could be helpful although the changes could be trivial. This is fine I think. The idea behind the releases so far is they've been a manual process. So we would only create releases that match the docker tags pushed and hence they would always match.

srenatus:
> We already create the OPA_VERSION-envoy-REV tag after each merge.

💡 I wasn't quite aware of the full picture then. So we've got a git tag and a dockerhub tag for every merge? I guess I don't understand where the git tag is created and pushed 🤔

srenatus:
No wait, I think I've got it now: On each merge, an image is pushed to dockerhub and tagged. The tags on github are manually created and pushed, they trigger the post-tag workflow that's drafting a release in github. That release is then published.

Correct? The tag pushed to github manually then **should match** the tag pushed to dockerhub already.

ashutosh-narkar:
> No wait, I think I've got it now: On each merge, an image is pushed to dockerhub and tagged. The tags on github are manually created and pushed, they trigger the post-tag workflow that's drafting a release in github. That release is then published.
> 
> Correct? The tag pushed to github manually then **should match** the tag pushed to dockerhub already.

Exactly !

srenatus:
So the question that still stands for me: when and to what should we update tags in our docs and quickstart yamls?

Based on this workflow, I'd think we could do that either
1. after a tag has been pushed to github: that would mean the docs-fixing commit is "one commit late" as far as the release is concerned (the tagged repo state would reference the previous release)
2. after each merge: that would cause gaps in the `-x` numbers, I think... we'll count commits-since-opa-bump, and we'll add commits fixing that... but it also wouldn't hurt 🤷 

With (2.), we'd still manually push the github tag, but the tag `x.y.z-envoy-#` would be applied to the docs-auto-update commit from the post-merge workflow.

Practically, it would mean we revert https://github.com/open-policy-agent/opa-envoy-plugin/commit/718f6459f7da4825af778417dada18daa446998b **and** fix the update script to follow the tagname logic from the `build/` script.

What do you think? (I'm quite sorry I haven't properly thought about this before when messing with the workflows here 😬)

ashutosh-narkar:
This is a workflow I propose:

1) On every merge to `main`, we publish an image to dockerhub. This includes versioned and `latest` tag images.
2) Run the doc update script in post-merge to update the versions in the README etc. We could simply update the docs, yamls, README etc. to use the `latest` tag which would mean we don't have to worry about this step.
3) Create an automatic release based on the tag used in `step 1` which would them trigger post-tag to create binaries etc. If we cannot automate tag creation as part of post-merge, then we do the manual way we currently do.

Does this flow make sense ? WDYT ?

srenatus:
>  2. [...] We could simply update the docs, yamls, README etc. to use the latest tag which would mean we don't have to worry about this step.

💯 I like this simplification. There's not too much value in tendering those tags in the quickstart/README if `envoy-latest` is just going to work.

With that, we can drop the extra-updater post-merge step.

> 3. Create an automatic release based on the tag used in step 1 which would them trigger post-tag to create binaries etc. If we cannot automate tag creation as part of post-merge, then we do the manual way we currently do.

We could do that, or we don't 😄  If we keep everything as-is, but change the readme/quickstart references, we could still have the "manual" release trigger (pushing a tag), which is nicer in my opinion. Let's release purposefully, not for every little merge.

ashutosh-narkar:
> Let's release purposefully, not for every little merge.

Sounds good to me !

srenatus:
@ashutosh-narkar I've included the cleanups in this PR. It would be great if you could have another look 👀  Thank you!

.github/workflows/checks.yaml
```diff
@@ -85,18 +79,14 @@ jobs:
...
           # Prepare quick_start.yaml
           make update-istio-quickstart-version
 
           # Make docker image available to k8s
           kind load docker-image openpolicyagent/opa:latest-istio
-        if: matrix.variant != 'tag'
 
       - name: Run e2e tests
         run: make test-e2e
-        timeout-minutes: 10
```
ashutosh-narkar, line 99: Is there some default here already that we don't need this anymore ?

ashutosh-narkar (approved):
Changes lgtm. Just one small comment. Can you please squash our commits before merging.

.github/workflows/checks.yaml
srenatus: Hmm why did I do that. 🤔 I'll revert and merge.

srenatus: Ah I remember. I haven't removed intentionally, it got removed reverting the other commit. I'll keep it instead! Good catch.

srenatus:
> Can you please squash our commits before merging.

Sure! I'd keep the OPA version bump separate, though, what do you think?

ashutosh-narkar:
> > Can you please squash our commits before merging.
> 
> Sure! I'd keep the OPA version bump separate, though, what do you think?

Sounds good. Thanks !

PR #270, opened 2021-07-25
envoyauth: Add support for parsing application/x-www-form-urlencoded request bodies

cbuto:
This commit adds support for parsing application/x-www-form-urlencoded request bodies into the parsed_body field. Prior to this change, the data was only available in the body or rawBody field in URL-encoded query string format.

Fixes open-policy-agent/opa#3600

Signed-off-by: Casey Buto <cbuto22@gmail.com>

### DIFF
--- a/envoyauth/request.go
+++ b/envoyauth/request.go
@@ -117,11 +117,11 @@ func getParsedBody(logEntry *logrus.Entry, headers map[string]string, body strin
 			}
 
 			if val, ok := headers["content-length"]; ok {
-				cl, err := strconv.ParseInt(val, 10, 64)
+				truncated, err := checkIfHTTPBodyTruncated(val, int64(len(body)))
 				if err != nil {
 					return nil, false, err
 				}
-				if cl != -1 && cl > int64(len(body)) {
+				if truncated {
 					return nil, true, nil
 				}
 			}
@@ -160,6 +160,27 @@ func getParsedBody(logEntry *logrus.Entry, headers map[string]string, body strin
 			if !known {
 				return nil, false, nil
 			}
+		} else if strings.Contains(val, "application/x-www-form-urlencoded") {
+			if body == "" {
+				return nil, false, nil
+			}
+
+			if val, ok := headers["content-length"]; ok {
+				truncated, err := checkIfHTTPBodyTruncated(val, int64(len(body)))
+				if err != nil {
+					return nil, false, err
+				}
+				if truncated {
+					return nil, true, nil
+				}
+			}
+
+			parsed, err := url.ParseQuery(body)
+			if err != nil {
+				return nil, false, err
+			}
+
+			data = map[string][]string(parsed)
 		} else {
 			logEntry.Debugf("content-type: %s parsing not supported", val)
 		}
@@ -243,4 +264,15 @@ func findMessageInputDesc(name string, svc protoreflect.ServiceDescriptor) (prot
 		return method.Input(), nil
 	}
 	return nil, fmt.Errorf("method %q not found", name)
+}
+
+func checkIfHTTPBodyTruncated(contentLength string, bodyLength int64) (bool, error) {
+	cl, err := strconv.ParseInt(contentLength, 10, 64)
+	if err != nil {
+		return false, err
+	}
+	if cl != -1 && cl > bodyLength {
+		return true, nil
+	}
+	return false, nil
 }
--- a/envoyauth/request_test.go
+++ b/envoyauth/request_test.go
@@ -184,10 +184,70 @@ func TestGetParsedBody(t *testing.T) {
 		}
 	  }`
 
+	requestContentTypeURLEncoded := `{
+		"attributes": {
+		  "request": {
+			"http": {
+			  "headers": {
+				"content-type": "application/x-www-form-urlencoded"
+			  },
+			  "body": "firstname=foo&lastname=bar"
+			}
+		  }
+		}
+	  }`
+
+	requestContentTypeURLEncodedMultipleKeys := `{
+		"attributes": {
+		  "request": {
+			"http": {
+			  "headers": {
+				"content-type": "application/x-www-form-urlencoded"
+			  },
+			  "body": "firstname=foo&lastname=bar&lastname=foobar"
+			}
+		  }
+		}
+	  }`
+
+	requestContentTypeURLEncodedTruncated := `{
+		"attributes": {
+		  "request": {
+			"http": {
+			  "headers": {
+				"content-type": "application/x-www-form-urlencoded",
+				"content-length": "1000"
+			  },
+			  "body": "firstname=foo&lastname=bar"
+			}
+		  }
+		}
+	  }`
+
+	requestContentTypeURLEncodedEmpty := `{
+		"attributes": {
+		  "request": {
+			"http": {
+			  "headers": {
+				"content-type": "application/x-www-form-urlencoded"
+			  },
+			  "body": ""
+			}
+		  }
+		}
+	  }`
 	expectedNumber := json.Number("42")
 	expectedObject := map[string]interface{}{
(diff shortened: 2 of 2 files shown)


envoyauth/request.go
```diff
@@ -160,6 +160,35 @@ func getParsedBody(logEntry *logrus.Entry, headers map[string]string, body strin
...
+
+			jsonBody, err := json.Marshal(parsed)
+			if err != nil {
+				return nil, false, err
+			}
+
+			err = util.UnmarshalJSON([]byte(jsonBody), &data)
+			if err != nil {
+				return nil, false, err
+			}
```
srenatus, line 191: 💭 Do we need the roundtrip? `parsed` is `url.Values`, which is an alias of `map[string][]string`. So, how about

```go
data = interface{}(parsed)
```
or something similar...? 🤔 

I believe we might even be able to say `data, err = url.ParseQuery(body)`, but I haven't tried that.

envoyauth/request_test.go
```diff
@@ -209,6 +252,9 @@ func TestGetParsedBody(t *testing.T) {
 		"content_type_json_with_special_chars": {input: createCheckRequest(requestBodyWithJSONSpecialChars), want: expectedJSONSpecialChars, isBodyTruncated: false, err: nil},
 		"empty_content":                        {input: createCheckRequest(requestEmptyContent), want: nil, isBodyTruncated: false, err: nil},
 		"body_truncated":                       {input: createCheckRequest(requestBodyTruncated), want: nil, isBodyTruncated: true, err: nil},
+		"content_type_url_encoded":             {input: createCheckRequest(requestContentTypeUrlEncoded), want: expectedUrlEncodedObject, isBodyTruncated: false, err: nil},
+		"content_type_url_encoded_empty":       {input: createCheckRequest(requestContentTypeUrlEncodedEmpty), want: nil, isBodyTruncated: false, err: nil},
+		"content_type_url_encoded_truncated":   {input: createCheckRequest(requestContentTypeUrlEncodedTruncated), want: nil, isBodyTruncated: true, err: nil},
```
srenatus, line 257: Let's add one with multiple values for one key, maybe? What do you think?

srenatus (commented):
This is a nice contribution, thanks for picking that up! 🎉 

Just one minor issue in the request handler 👇 🙃

envoyauth/request.go
cbuto: @srenatus I think you are right, we can just return the `parsed` data as a `url.Values` type and use that same type for object we are asserting with. The initial reason I went with converting to JSON and back was to not use the `url.Values` type in the tests but I think its fine! I pushed a fix in f76cbd6!

envoyauth/request_test.go
cbuto: good call, fixed in f76cbd6

envoyauth/request.go
srenatus: > The initial reason I went with converting to JSON and back was to not use the url.Values type in the tests but I think its fine! 

Hm! I guess if we turn `url.Values` back into the native type, i.e. `data = parsed.(map[string][]string)`, we wouldn't have to do that, would we? What do you think?

cbuto: I poked at this for a bit but I think the only way to convert the `url.Values` type back to its native type is by looping over the values and copying them one by one into a new variable with the desired type, which might be better then doing thee marshal/unmarshal dance. 

Here's an example: https://play.golang.org/p/jZJ-J7bTUGk

cbuto: Following up, I actually ran a benchmark (just on my laptop) with the two options and it looks like looping through the values and copying it into another variable ended up being faster. Here's the [code](https://play.golang.org/p/ntlkHApdK65) (the playground won't work 😄 )

```
goos: darwin
goarch: amd64
pkg: example
cpu: Intel(R) Core(TM) i7-5650U CPU @ 2.20GHz
BenchmarkLoop-4          1691784               831.3 ns/op
BenchmarkMarshal-4        175098              6107 ns/op
PASS
ok      example 3.472s
```

srenatus: My bad. Could you try `data = map[string][]string(parsed)`, please? I'd suspect it to be quicker, too.

Thanks for bearing with me! 😅

cbuto: ah yeah🤦 , its a definitely bit faster and cleaner. Got a little side tracked on all the possibilities 👨‍🔬 . Fixed in 9314ee5!

srenatus (approved):
The code looks good to me. Thanks for contributing. 🎉 

I haven't run this myself with an instance of Envoy and an actual curl req with this input -- have you? 

Finally, now is a good time to squash those commits 🚀 Thanks!

srenatus:
Oh, the second and third commits don't have a message, so I'd be OK to _squash and merge_ them myself using github's button. ✔️

cbuto:
> The code looks good to me. Thanks for contributing. 🎉
> 
> I haven't run this myself with an instance of Envoy and an actual curl req with this input -- have you?
> 
> Finally, now is a good time to squash those commits 🚀 Thanks!

🎉  I ran this with Envoy, here's what the decision log ended up looking like:

```
{
  "decision_id": "412cef1a-0017-4122-ad33-173d32f620fd",
  "input": {
    "attributes": {
      "destination": {
        "address": {
          "socketAddress": {
            "address": "172.18.0.4",
            "portValue": 51051
          }
        }
      },
      "metadataContext": {},
      "request": {
        "http": {
          "body": "param1=value1&param2=value2",
          "headers": {
            ":authority": "127.0.0.1:51051",
            ":method": "POST",
            ":path": "/",
            ":scheme": "http",
            "accept": "*/*",
            "content-length": "27",
            "content-type": "application/x-www-form-urlencoded",
            "user-agent": "curl/7.64.1",
            "x-envoy-auth-partial-body": "false",
            "x-forwarded-proto": "http",
            "x-request-id": "4c8da78f-ab2e-49f9-8cfe-dc79e5d040a2"
          },
          "host": "127.0.0.1:51051",
          "id": "2047903939553692896",
          "method": "POST",
          "path": "/",
          "protocol": "HTTP/1.1",
          "scheme": "http",
          "size": "27"
        },
        "time": "2021-07-29T20:38:47.053771Z"
      },
      "source": {
        "address": {
          "socketAddress": {
            "address": "172.18.0.1",
            "portValue": 57098
          }
        }
      }
    },
    "parsed_body": {
      "param1": [
        "value1"
      ],
      "param2": [
        "value2"
      ]
    },
    "parsed_path": [
      ""
    ],
    "parsed_query": {},
    "truncated_body": false,
    "version": {
      "encoding": "protojson",
      "ext_authz": "v3"
    }
  },
  "labels": {
    "id": "f608cc28-8288-4ff1-8bb9-4676c2cb8672",
    "version": "0.30.2-envoy-4"
  },
  "level": "info",
  "metrics": {
    "timer_rego_query_compile_ns": 239100,
    "timer_rego_query_eval_ns": 108700,
    "timer_server_handler_ns": 1012900
  },
  "msg": "Decision Log",
  "path": "envoy/authz/allow",
  "requested_by": "",
  "result": false,
  "time": "2021-07-29T20:38:47Z",
  "timestamp": "2021-07-29T20:38:47.0608211Z",
  "type": "openpolicyagent.org/decision_logs"
}
```

I squashed the commits in e6b6099 just in case!

srenatus:
Uhm I'm really confused what's happening in the github actions here. I'm poking around, it seems like it's working with the latest versions of istio and kind (#272), but I don't understand why it should fail with this PR. I'm having another look.

srenatus:
Could you please rebase this? Maybe it'll pass tests then.

cbuto:
> Could you please rebase this? Maybe it'll pass tests then.

done! 🤞

ashutosh-narkar:
Thanks for the contribution @cbuto !

PR #272, opened 2021-07-30
remove unused make target, bump istio+kind versions

srenatus:
I've not been able to get a grip on what happens with the istio e2e tests in #273 (and #270), but it looks like bumping dependencies helps? So let's do that. 🐑

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -79,9 +79,6 @@ jobs:
           sudo apt-get update -y
           sudo apt-get install -y bats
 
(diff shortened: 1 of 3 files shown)


ashutosh-narkar (approved):

ashutosh-narkar:
@srenatus did you check if updating the versions, results in passing tests in #270  ?

srenatus:
@ashutosh-narkar I've run an empty PR and it had the same issues, resolved here... So while I don't fully understand what's going on, if the issue is gone, I'm ok with it...

PR #274, opened 2021-08-06
workflow: let dependabot handle all go mod dependencies

srenatus:
Previously, we've only used it for OPA.

~🚧 We'll have to update our script logic if a dependabot-authored commit no longer signifies that OPA was updated.~ #300 resolved that.


srenatus:
TBC

srenatus:
Let's see how it goes 🤞

PR #277, opened 2021-08-12
Use CUSTOM AuthorizationPolicies in Istio example

bburky:
I tested Istio 1.9+'s new CUSTOM AuthorizationPolicy feature using injected OPA sidecars like the this project's Istio example and got it to work. This replaces the EnvoyFilter with a "provider" configuration entry in the mesh config which is referenced by the AuthorizationPolicy. To easily allow configuring the mesh config, I switched the Istio installation method to use the operator.

I think this is the new recommended way to do external authorization with Istio, so I'm submitting this PR if you'd like to update the Istio example.

See Istio documentation for more information:
https://istio.io/latest/docs/tasks/security/authorization/authz-custom/
https://istio.io/latest/blog/2021/better-external-authz/

* * * 

Minimum Istio version increased to 1.9. Installation instructions use Istio operator to allow configuring an envoyExtAuthzGrpc extension provider in the mesh config.

Localhost ServiceEntry added to allow provider to target a sidecar (127.0.0.1 IP address) instead of a Service.

A CUSTOM AuthorizationPolicy matches workloads and applies policy using the external authorization provider.

EnvoyFilter is removed.

### DIFF
--- a/build/install-istio-with-kind.sh
+++ b/build/install-istio-with-kind.sh
@@ -7,7 +7,7 @@ set -x
 GOARCH=$(go env GOARCH)
 GOOS=$(go env GOOS)
 KIND_VERSION=0.11.1
-ISTIO_VERSION=1.8.6
+ISTIO_VERSION=1.9.7
 
 # Download and install kind
 curl -L https://github.com/kubernetes-sigs/kind/releases/download/v${KIND_VERSION}/kind-${GOOS}-${GOARCH} --output kind && chmod +x kind && sudo mv kind /usr/local/bin/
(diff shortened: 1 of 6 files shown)


bburky:
Not quite sure why the Istio end-to-end tests are timing out. I ran the bats tests locally and they seem to work.

ashutosh-narkar:
@bburky thanks for this contribution! Given that the `CUSTOM` action is currently an experimental feature what do you think about keeping the existing example which uses the `EnvoyFilter` as is and then adding a folder under `examples` with this new setup ?

Once `CUSTOM` action becomes more stable, we can deprecate the `EnvoyFilter` example.

ashutosh-narkar:
Closing this for now. We can revisit this when [CUSTOM](https://istio.io/latest/docs/tasks/security/authorization/authz-custom/) action becomes more stable.

bburky:
Sorry for not updating the PR. I wasn't sure how to avoid too much code duplication and didn't finish updating it.

Closing this PR is fine. You can use this code for reference once CUSTOM is stabilized.

PR #278, opened 2021-08-23
fix deadlock bundle decision plugins

srenatus:
🚧 The txn is definitely not passed to the decision logger.

It now is, but I haven't verified the behaviour in any real way yet. It would be nice to reproduce the dead lock via a test, or the race detector.

Fixes https://github.com/open-policy-agent/opa/issues/3722.


ashutosh-narkar:
Closing this in favor of https://github.com/open-policy-agent/opa-envoy-plugin/pull/279.

PR #282, opened 2021-10-14
feat(envoyauth_request): add partial support for multipart/form-data

emaincourt:
Hi,

At the moment, `opa-envoy-plugin` is not able to parse `multipart/form-data` payloads. I don't know whether this was made on purpose because of potential issues regarding the parsing of uploaded files or for any other reason so please let me know it all that sounds wrong to you folks.

As far as I am aware of, Envoy needs to have `pack_as_bytes` set to `true` to be able to process binary payloads such as files. That's why I'm evaluating both cases where `body` or `rawBody` could be empty, depending on the configuration of the `EnvoyFilter`.

Also, I've been thinking about only reading form parts that have an explicit `Content-Type` header set to `text/plain` and `application/json` to reduce parsing overhead and memory allocation. However, most of the times `text/plain` parts themselves are missing this header resulting in basically no fields being parsed at all.

Parsing returns a `map[string][]interface{}` as multiple parts can share the same `name` property.

Unit tests have been added and I've ran this image on top of an existing cluster to perform some basic testing in both cases where `body` or `rawBody` would be empty depending on the `ext_authz` filter configuration.

What do you folks think ?

Thanks in advance for taking the time to review my proposal.

### DIFF
--- a/envoyauth/request.go
+++ b/envoyauth/request.go
@@ -4,6 +4,9 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
+	"io"
+	"mime"
+	"mime/multipart"
 	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


emaincourt:
@ashutosh-narkar Thanks for the review. I addressed all your comments.

PR #291, opened 2021-11-05
examples: updated the README files to reflect kubernetes version

olamiko:
Fixes open-policy-agent/opa#3906

Signed-off-by: Olamide Omolola <omololaolamidex@gmail.com>


anderseknert (approved):
LGTM! Would be good to know why we can't use v1 for admissionReviewVersions, but we can look into that later as this is still an improvement.

PR #295, opened 2021-11-17
support parsed_body input for json payloads sent as bytes (pack_as_byes: true) #4027

vinhph0906:
Following the discussion at  [open-policy-agent/opa#4027](https://github.com/open-policy-agent/opa/issues/4027), I have a quick change update to support this issue.


srenatus:
Thanks for looking into this! As mentioned in the issue, it would be nice if we could add test case here.

Also the rawBody is base64-encoded, I believe we would need to decode it so it can be unmarshalled into JSON and returned from that function.

srenatus:
👉 https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/envoyauth/request_test.go#L43 An extra case here would do the trick, if you'd like to add it.

vinhph0906:
In my first time looking into decision logs, I think rawBody is base64 encoded too.  But after the change and debug, I realize rawBoby just is []byte(body) and base64 encoded we saw in logs just cause by https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/envoyauth/request.go#L43 and https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/vendor/google.golang.org/protobuf/encoding/protojson/encode.go#L288 
``

srenatus:
Ah thanks for helping me see that. Let's add a simple test case anyways, please!

PR #296, opened 2021-11-18
support parsed_body input for json payloads sent as bytes (pack_as_byes: true) #4027

vinhph0906:
Following the pull request at #295, I have updated the test case for this case. But it is my first contribution project and something went wrong when I try to push. So I retry in this pull request.


srenatus (approved):
Thanks a lot, that's a nice first contribution! 👏 🎉

PR #299, opened 2021-12-01
envoyauth: enable print() functionality

srenatus:
Fixes https://github.com/open-policy-agent/opa/issues/3951.

### DIFF
--- a/envoyauth/evaluation.go
+++ b/envoyauth/evaluation.go
@@ -1,6 +1,7 @@
 package envoyauth
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"sync"
@@ -10,6 +11,7 @@ import (
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/storage"
+	"github.com/open-policy-agent/opa/topdown"
 	iCache "github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/sirupsen/logrus"
 )
@@ -27,7 +29,7 @@ type EvalContext interface {
 	SetPreparedQuery(*rego.PreparedEvalQuery)
 }
 
-//Eval - Evaluates an input against a provided EvalContext and yields result
+// Eval - Evaluates an input against a provided EvalContext and yields result
 func Eval(ctx context.Context, evalContext EvalContext, input ast.Value, result *EvalResult, opts ...func(*rego.Rego)) error {
 	var err error
 
@@ -61,6 +63,9 @@ func Eval(ctx context.Context, evalContext EvalContext, input ast.Value, result 
 		return err
 	}
 
+	var buf bytes.Buffer
+	ph := topdown.NewPrintHook(&buf)
+
 	var rs rego.ResultSet
 	rs, err = evalContext.PreparedQuery().Eval(
 		ctx,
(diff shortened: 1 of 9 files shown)


envoyauth/evaluation.go
```diff
@@ -61,43 +63,49 @@ func Eval(ctx context.Context, evalContext EvalContext, input ast.Value, result
 		return err
 	}
 
+	var buf bytes.Buffer
+	ph := topdown.NewPrintHook(&buf)
```
tsandall, line 67: Any particular reason we're using a bytes.Buffer here instead of providing a print hook that sends directly to logrus?

tsandall: As I wrote this comment, I realized we should update the plugin to use the logging package in OPA so that the plugin is not directly dependent on logrus.

tsandall (approved):
Looks fine. I wonder if improving the print hook usage would be worthwhile but as long as the print() calls are automatically erased in prod deployments when --log-level=error, I think we're fine.

envoyauth/evaluation.go
srenatus: > Any particular reason we're using a bytes.Buffer here instead of providing a print hook that sends directly to logrus?

I'm trying that now -- I had just taken the route of adding to `EvalResult` first, and then it made sense to do it this way. Let's see where the other path goes.

> As I wrote this comment, I realized we should update the plugin to use the logging package in OPA so that the plugin is not directly dependent on logrus.

Agreed, let's make that a follow-up.

srenatus:
OK I've tried directly hooking up the logger. It works nicely. But it's not as simple to test, and I've thought that introducing some mockery for logrus when it's going to be ditched in the next PR isn't a sensible use of our time 😅 

WDYT, let's merge this as-is and take on the logging refactoring?

srenatus:
Included the logging refactoring here. I think there's no way to make it 100% backwards-compatible, but it's not too bad either.

srenatus:
I haven't double checked yet that the log level settings still work. Will do that before merging.

envoyauth/evaluation.go
```diff
@@ -27,16 +28,16 @@ type EvalContext interface {
 	SetPreparedQuery(*rego.PreparedEvalQuery)
 }
 
-//Eval - Evaluates an input against a provided EvalContext and yields result
-func Eval(ctx context.Context, evalContext EvalContext, input ast.Value, result *EvalResult, opts ...func(*rego.Rego)) error {
+// Eval - Evaluates an input against a provided EvalContext and yields result
+func Eval(ctx context.Context, logger logging.Logger, evalContext EvalContext, input ast.Value, result *EvalResult, opts ...func(*rego.Rego)) error {
```
tsandall, line 32: side note: This is a good example of why we want to use structs or functional options for arguments... we can't extend this API without breaking callers.

tsandall (approved):
@srenatus assuming you've tested these changes and verified the print() calls are surprised at ERROR, LGTM. @mjungsbluth we've got some very small backwards incompatible changes that are being made to the envoyauth package. Apologies for the inconvenience.

envoyauth/evaluation.go
srenatus: I guess if we have to break it, we could break it in a way that gets us into a better place...

mjungsbluth: Alternatively you could have added the logger to EvalContext. I will update our usage after this is released.

srenatus: That's true, but not much less backward-compatibility-breaking, is it? Extending the EvalContext interface, I mean. But  I'll play with that a bit, and see if it ends up neater anyways.

srenatus:
✔️ Works with logging: level=debug gives us much output, info a middle ground including print()s, error not even print().

✔️ Also moved the logging.Logger into the EvalContext, thanks for the suggestion

PR #305, opened 2021-12-07
Failure to parse request should result in a bad request response.

ChezCrawford:
A potential solution to https://github.com/open-policy-agent/opa/issues/4101

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -317,7 +317,22 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 
 	input, err = envoyauth.RequestToInput(req, logger, p.cfg.protoSet)
 	if err != nil {
-		return nil, stop, err
+		logger.WithFields(map[string]interface{}{"err": err}).Info("Unable to parse request.")
+
+		deniedResponse := &ext_authz_v3.DeniedHttpResponse{
+			Status: &ext_type_v3.HttpStatus{
+				Code: ext_type_v3.StatusCode(ext_type_v3.StatusCode_BadRequest),
+			},
+		}
+
+		resp := &ext_authz_v3.CheckResponse{
+			HttpResponse: &ext_authz_v3.CheckResponse_DeniedResponse{
+				DeniedResponse: deniedResponse,
+			},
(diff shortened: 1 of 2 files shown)


ChezCrawford:
I have two outstanding questions about this change:

1) Should we make this behavior "opt-in" via some sort of configuration?

2) Should do any additional logging for this type of response?

srenatus:
Thanks for picking this up 🎉 

> Should we make this behavior "opt-in" via some sort of configuration?

I suppose one new configurable we could introduce is "error on bad input" and it could default to StatusCode_BadRequest. But if you somehow depend on the other behaviour, or want this to be 403, you could configure it accordingly...? 🤔 

> Should do any additional logging for this type of response?

What would you have in mind there? The error is logged, so that seems OK.

internal/internal.go
```diff
@@ -308,7 +308,22 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
...
+			Status: &ext_type_v3.HttpStatus{
+				Code: ext_type_v3.StatusCode(ext_type_v3.StatusCode_BadRequest),
+			},
+		}
+
+		resp := &ext_authz_v3.CheckResponse{
+			HttpResponse: &ext_authz_v3.CheckResponse_DeniedResponse{
+				DeniedResponse: deniedResponse,
+			},
+			Status: &rpc_status.Status{Code: int32(code.Code_INVALID_ARGUMENT)},
```
srenatus, line 323: It just now strikes me as odd that you've got to define both the http status and this one. 🤔 

From the docs, the inner one would overwrite the outer one. So let's only specify one of the two? This would also make it simpler to configure this.

srenatus:
Sorry for the radio silence. I'll get back to this next week!

srenatus:
> If not, I am happy to add one of the config options above and default the behavior to returning 403 as it works today...

I think this is preferable. Everyone is happiest if they can just upgrade without any seconds thoughts, and that's what we'd need to do to achieve this.

Sorry it took me so long to come up with that 😅

srenatus:
Let's reopen when you get a chance to get back to this. 😃

PR #323, opened 2022-03-16
feat(grpc): allow to configure grpc msg rcv/send sizes

emaincourt:
Hi,

I know that this has already been opened quite a few times without never being released so I'll try. This change introduces both `grpc-max-rcv-message-size` and `grpc-max-send-message-size` with default values being set to the ones from the official `grpc-go` library. It helps when dealing with large payloads.

I've deployed it on my clusters and it seems to work as expected.

Thanks in advance for reading

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -7,6 +7,7 @@ package internal
 import (
 	"context"
 	"fmt"
+	"math"
 	"net"
 	"net/url"
 	"os"
@@ -47,6 +48,11 @@ const defaultPath = "envoy/authz/allow"
 const defaultDryRun = false
 const defaultEnableReflection = false
 
+// Those are the defaults from grpc-go.
+// See https://github.com/grpc/grpc-go/blob/master/server.go#L58 for more details.
+const defaultGRPCServerMaxReceiveMessageSize = 1024 * 1024 * 4
+const defaultGRPCServerMaxSendMessageSize = math.MaxInt32
+
 // PluginName is the name to register with the OPA plugin manager
 const PluginName = "envoy_ext_authz_grpc"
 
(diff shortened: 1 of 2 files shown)


ashutosh-narkar:
Thanks for the contribution. It would be helpful if we exposed this via the plugin config and then fallback to defaults if not set.

emaincourt:
Hi @ashutosh-narkar,

Thanks for reviewing the PR. Actually I'm not sure to understand. At the moment it is configurable by using flags such as follows:

```
- --set=plugins.envoy_ext_authz_grpc.grpc-max-recv-msg-size=<uint32>
- --set=plugins.envoy_ext_authz_grpc.grpc-max-send-msg-size=<uint32>
```

If not provided, default values will be used. Where else should I update the code to match your requirements ?

Thanks in advance

ashutosh-narkar:
Can you please add some unit tests ? Also we'll have to update the [docs](https://github.com/open-policy-agent/opa/blob/main/docs/content/envoy-introduction.md) once this change is in.

emaincourt:
Hi @ashutosh-narkar,

I've added the tests and opened this draft PR for updating the documentation: https://github.com/open-policy-agent/opa/pull/4459.

Let me know what you think about them.

Thanks in advance.

ashutosh-narkar:
@emaincourt this looks good. Thanks for this contribution. Can you please squash your commits and rebase and then we can merge this.

emaincourt:
@ashutosh-narkar Should be all good. Thanks for the review.

ashutosh-narkar (approved):
LGTM

ashutosh-narkar:
@emaincourt can you please work on updating https://github.com/open-policy-agent/opa/pull/4459 ? Thanks

PR #324, opened 2022-03-21
Support to control headers for the upstream and client

ashutosh-narkar:
This commit updates the policy interface to allow
a policy author to define headers that should be
removed before a request is sent upstream and added
to the response sent to the client. Both these
are applicable to successful ext authz responses.

Fixes: open-policy-agent/opa/issues/4417

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -103,56 +103,61 @@ func (result *EvalResult) IsAllowed() (bool, error) {
 	return false, result.invalidDecisionErr()
 }
 
+// GetRequestHTTPHeadersToRemove - returns the http headers to remove from the original request before dispatching
+// it to the upstream
+func (result *EvalResult) GetRequestHTTPHeadersToRemove() ([]string, error) {
+	headersToRemove := []string{}
+
+	switch decision := result.Decision.(type) {
+	case bool:
+		return headersToRemove, nil
+	case map[string]interface{}:
+		var ok bool
+		var val interface{}
+
+		if val, ok = decision["request_headers_to_remove"]; !ok {
+			return headersToRemove, nil
+		}
+
+		switch val := val.(type) {
+		case []string:
+			return val, nil
+		case []interface{}:
+			for _, vval := range val {
+				header, ok := vval.(string)
+				if !ok {
+					return nil, fmt.Errorf("type assertion error")
+				}
+
+				headersToRemove = append(headersToRemove, header)
+			}
+			return headersToRemove, nil
+		default:
+			return nil, fmt.Errorf("type assertion error")
+		}
+	}
+
+	return nil, result.invalidDecisionErr()
+}
+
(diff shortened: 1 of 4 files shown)


envoyauth/response.go
```diff
@@ -281,3 +299,64 @@ func (result *EvalResult) GetResponseEnvoyHTTPStatus() (*ext_type_v3.HttpStatus,
...
+	takeResponseHeaders := func(headers map[string]interface{}, targetHeaders *http.Header) error {
+		for key, value := range headers {
+			var headerVal string
+			var ok bool
+			if headerVal, ok = value.(string); !ok {
+				return fmt.Errorf("type assertion error")
+			}
+
+			targetHeaders.Add(key, headerVal)
+		}
```
srenatus, line 314: Just trying out an inversion here

```suggestion
		for key, value := range headers {
			if headerVal, ok := value.(string); ok {
				targetHeaders.Add(key, headerVal)
				continue
			}
			return fmt.Errorf("type assertion error")
		}
```

🤷 one seems as good as the other. Let's keep it.

internal/internal_test.go
```diff
@@ -1017,6 +1017,110 @@ func TestConfigWithProtoDescriptor(t *testing.T) {
...
+	}
+
+	response := output.GetOkResponse()
+	if response == nil {
+		t.Fatal("Expected OkHttpResponse struct but got nil")
+	}
+
+	headers := response.GetHeadersToRemove()
+	if len(headers) != 2 {
+		t.Fatalf("Expected two headers but got %v", len(headers))
```
srenatus, line 1062: [nit] 🤔 I'd use `t.Fatal[f]` only when the rest of the test body is entirely impossible without this check.  I don't suppose that's the case here.

srenatus (approved):
Code looks good. Nitpicks aside,

- Do we have any place to document this in _this_ repo, or is documentation entirely happening in the OPA website docs for this enhancement?
- I genuinely can't remember: are there implications for v2 vs v3 of the transport API? I think I only see v3 in the change; but I also don't remember much about automatic v3 -> v2 stuff we do. We we cool with v2? 💭

ashutosh-narkar:
> Code looks good. Nitpicks aside,
> 
> * Do we have any place to document this in _this_ repo, or is documentation entirely happening in the OPA website docs for this enhancement?

I will add a follow-up PR in the OPA docs for this.


> * I genuinely can't remember: are there implications for v2 vs v3 of the transport API? I think I only see v3 in the change; but I also don't remember much about automatic v3 -> v2 stuff we do. We we cool with v2? 💭

We should be ok with the V3 to V2 response conversion. We don't use the new V3 fields to generate the V2 response. At some point we should remove the V2 stuff. Envoy v1.18+ does not support it anyway so we should look into removing this as well. Envoy v1.18 was released almost a year ago.

srenatus:
Thanks!

PR #328, opened 2022-04-05
proxy_init: migrate proxy_init container

charlesdaniels:
Move the `proxy_init` container over from the contrib repository.

### DIFF
--- a/.github/workflows/proxy_init-tag.yaml
+++ b/.github/workflows/proxy_init-tag.yaml
@@ -1,0 +1,37 @@
+name: proxy_init Tag
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'proxy_init-v*'
+
+jobs:
+  deploy:
+    name: Build and deploy proxy_init container
+    steps:
+
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set TAG_NAME in Environment
+        # Subsequent jobs will be have the computed tag name
+        run: echo "TAG_NAME=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Workaround Buildx Errors
+        run: build/buildx_workaround.sh
+
+      - name: Login to Docker
+        run: make docker-login
+        env:
+          DOCKER_USER: ${{ secrets.DOCKER_USER }}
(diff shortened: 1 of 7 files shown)


.github/workflows/proxy_init-tag.yaml
```diff
@@ -0,0 +1,37 @@
+name: proxy_init Tag
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'proxy_init-v*'
```
srenatus, line 8: clever

proxy_init/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:xenial
```
srenatus, line 1: 🤔 This will make for a large image giving this little functionality. But it's OK if that's what we had before. Could be a future optimization.

proxy_init/Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+ADD ./proxy_init.sh /proxy_init.sh
+RUN chmod 755 /proxy_init.sh
```
srenatus, line 4: I think I've read somewhere you can do this now:
```suggestion
ADD --chmod=0755 ./proxy_init.sh /proxy_init.sh
```

But I might be wrong 🤔

srenatus (approved):
Thanks for looking into this! LGTM. But I might be missing context re: this being a draft still. What are we waiting for?

charlesdaniels:
> Thanks for looking into this! LGTM. But I might be missing context re: this being a draft still. What are we waiting for?

A big part of this is getting it to build and push images to Docker Hub automatically, which I think I have working code for, but I'm not certain how to test, so I was hoping someone else would have ideas on that front.

srenatus:
> A big part of this is getting it to build and push images to Docker Hub automatically, which I think I have working code for, but I'm not certain how to test, so I was hoping someone else would have ideas on that front.

Let's try? I'm down to merging this and seeing what happens 😎

charlesdaniels:
Ok, we can give it a try then...

proxy_init/Dockerfile
charlesdaniels: Agreed, I usually go with Alpine based images, but I didn't want to spent a ton of time debugging if switching was to break things. One of the slim Debian images might be a good choice in the future.

proxy_init/Dockerfile
charlesdaniels: https://docs.docker.com/engine/reference/builder/#add

I think you are mixed up with `--chown`?

PR #329, opened 2022-04-24
Build infra scaffolding

c0c0n3:


c0c0n3:
sorry I made a mistake, ignore this PR, I was just playing around in my fork and accidentally submitted a PR.
sorry about that!

srenatus:
👀 Totally caught my interest here. What are you about to do? 😃

c0c0n3:
Hi @srenatus!

That PR was just to get my local dev env going---I use [Nix][nix] to manage my deps. Also, I was thinking of packaging your excellent plugin for Nix since there's a package for OPA but it doesn't look like there's one for your plugin.

My main goal is to explore what it'd take to bolt OIDC on top of your plugin. Your code is very well written so it shouldn't be hard, but as usual the devil is in the details. We're developing a cloud platform [where we use your plugin][sec] to secure our resources. Your plugin works great for service to service communication but we need to be able to handle end user interaction too where if the user isn't logged in, they should go through an OAuth2 flow (Auth Code) to get authenticated and then authorised by our OPA policies when redirected back to the resource they originally requested.

There are solutions out there like OAuth2 Proxy and the Envoy OAuth2 filter that could help with our use case but we'd need to extend them since we've got to be able to handle multiple tenants within the same Auth Server---Keycloak at the moment. We'd like to be able to do something similar to this

1. User accesses resource `R` w/ no credentials
2. OPA policy denies access, but returns an indication the user should be redirected to the login page of a given tenant `T`.
3. The plugin extension we're thinking of writing would then carry out an Auth Code flow, stick the token in an HTTP-only cookie, call the policy again and finally redirect the browser to `R` if the policy allows it.
4. For any subsequent request, the extension would translate the cookie into an authorization header before evaluating the policy, and ideally also refresh the token if it's close to its expiry---that way active users won't have to keep on login in.

The reason we can't use OAuth2 Proxy or the Envoy OAuth2 filter is that `T` is a function of `R` so the OAuth2 URL to use to get a token is in general a function of the request too. E.g.

1. Eve is a user with tenant `t1`. She does a `GET /t1/bananas` w/o credentials. She should go login at `/auth/realms/t1`.
2. Dan is a user with tenant `t2`. He does a `GET /t2/bananas` w/o credentials. He should go login at `/auth/realms/t2`.

The reason we'd like to be able to do this is that ideally we should use a Keycloak realm for each tenant.

My thinking at the moment is that extending your plugin should be easier than the others. Also it'd come with the extra benefit of reduced latency since we won't have to deploy an additional service.
 



[sec]: https://github.com/c0c0n3/kitt4sme.live/tree/main/deployment/mesh-infra/security
[nix]: https://nixos.org/

srenatus:
Thanks for sharing! This looks reasonable on a high level -- what I'm not sure about is what's lacking in the plugin for your use case? I think custom return codes are possible via object responses... but I might be missing something. Perhaps you could go into detail a bit please on what your planned extension would do, please? 👀

c0c0n3:
> what your planned extension would do, please?

hey, hey, yes, here's some more details about what I'd like to do. Basically, I'd like to be able to use OPA Plugin both as a policy enforcement point through Envoy (current functionality) and as an OIDC flow manager---similar to OAuth2 proxy or Envoy OAuth2 filter, this functionality isn't there and that's what I'm working on.

Here's a diagram of what I could do to have OPA Plugin carry out the [OAuth2 code flow][code-grant].

```
+---------+                                   +-------+                                    +-----------+                                    +-------------+ +---------+
| Browser |                                   | Envoy |                                    | OPAPlugin |                                    | AuthServer  | | Service |
+---------+                                   +-------+                                    +-----------+                                    +-------------+ +---------+
     |                                            |                                              |                                                 |             |
     | GET /tenantA/resource                      |                                              |                                                 |             |
     |------------------------------------------->|                                              |                                                 |             |
     |                                            |                                              |                                                 |             |
     |                                            | check(get, /tenantA/resource)                |                                                 |             |
     |                                            |--------------------------------------------->|                                                 |             |
     |                                            |                                              |                                                 |             |
     |                                            |        no token, redirect to tenantA's login |                                                 |             |
     |                                            |<---------------------------------------------|                                                 |             |
     |                                            |                                              |                                                 |             |
     |              401, Loc: /auth/tenantA/login |                                              |                                                 |             |
     |<-------------------------------------------|                                              |                                                 |             |
     |                                            |                                              |                                                 |             |
     | GET /auth/tenantA/login                    |                                              |                                                 |             |
     |-------------------------------------------------------------------------------------------------------------------------------------------->|             |
     |                                            |                                              |                                                 |             |
     |                                            |                                              |                   return login form for tenantA |             |
     |<--------------------------------------------------------------------------------------------------------------------------------------------|             |
     |                                            |                                              |                                                 |             |
     | POST /auth/tenantA/login { user's credentials }                                           |                                                 |             |
     |-------------------------------------------------------------------------------------------------------------------------------------------->|             |
     |                                            |                                              |                                                 |             |
     |                                            |                                              |                        303, Loc: /opa?code=c0d3 |             |
     |<--------------------------------------------------------------------------------------------------------------------------------------------|             |
     |                                            |                                              |                                                 |             |
     | GET /opa?code=c0d3                         |                                              |                                                 |             |
     |------------------------------------------------------------------------------------------>|                                                 |             |
     |                                            |                                              |                                                 |             |
     |                                            |                                              | POST /auth/tenantA/token { code = c0d3 }        |             |
     |                                            |                                              |------------------------------------------------>|             |
     |                                            |                                              |                                                 |             |
     |                                            |                                              |                             { token = t0.k3.n } |             |
     |                                            |                                              |<------------------------------------------------|             |
     |                                            |                                              |                                                 |             |
     |                                            | 303, Loc: /tenantA/resource, Cookie: t0.k3.n |                                                 |             |
     |<------------------------------------------------------------------------------------------|                                                 |             |
     |                                            |                                              |                                                 |             |
     | GET /tenantA/resource, Cookie: t0.k3.n     |                                              |                                                 |             |
     |------------------------------------------->|                                              |                                                 |             |
     |                                            |                                              |                                                 |             |
     |                                            | check(get, /tenantA/resource, t0.k3.n)       |                                                 |             |
     |                                            |--------------------------------------------->|                                                 |             |
     |                                            |                                              |                                                 |             |
     |                                            |        ok, valid token, policy allows access |                                                 |             |
     |                                            |<---------------------------------------------|                                                 |             |
     |                                            |                                              |                                                 |             |
     |                                            | GET /tenantA/resource                        |                                                 |             |
     |                                            |------------------------------------------------------------------------------------------------------------->|
     |                                            |                                              |                                                 |             |
     |                                            |                                              |                                        200, { resource data } |
     |                                            |<-------------------------------------------------------------------------------------------------------------|
     |                                            |                                              |                                                 |             |
     |                    200, { resource data }  |                                              |                                                 |             |
     |<-------------------------------------------|                                              |                                                 |             |
     |                                            |                                              |                                                 |             |
```

My understanding is that as OPA Plugin stands at the moment, I can easily get to redirect the browser (message 4 in the diagram, `401, Loc: /auth/tenantA/login`) but then I'd need another service e.g. OAuth2 proxy to carry out the rest of the flow.


[code-grant]: https://github.com/c0c0n3/platos-cave/blob/master/cloud-computing/iam/oidc/oauth2-distilled.md#authorisation-code-grant

ashutosh-narkar:
@c0c0n3 this looks cool! I would imagine you having a service that does the OIDC flow and then using the token from that and pass it on to OPA. Something like [this](https://github.com/StyraInc/opa-apigee-demo).  It's not clear to me how the browser request is sent to Envoy.

c0c0n3:
@ashutosh-narkar hey don't get too excited about it. I'm just hacking the software to pieces to make it work for my use case :-) If I realise I can make it more general and useful to others I'll submit a PR to you guys :-)

> I would imagine you having a service that does the OIDC flow and then using the token from that and pass it on to OPA. 

Yep. That was my initial set up with the new Envoy OAuth2 filter managing the OIDC bit.

> Something like [this](https://github.com/StyraInc/opa-apigee-demo).

Similar but my goal is **not** to have each service manage the OIDC flow. It looks like the Front End lambda in the diagram is  [managing itself the OIDC flow][front-end.oidc]. While this is usually a reasonable choice, in our cloud platform we've got to deal with a potentially large number of legacy services that would be a mission to retrofit and make them do the OIDC flow. My idea is to implement the OIDC flow once and for all in an external service so that the other services in the mesh don't have to do that. Like I said there's already excellent solutions out there like OAuth2 proxy that can help, but in my case I need to be able to manage multi-tenancy as I mentioned in my comments early. Since we're already (very happy!) OPA Plugin users, I though extending it with OIDC flow management would be the best option for us.

> It's not clear to me how the browser request is sent to Envoy.

Yah, good point. I got a bit too much detail out of the diagram I guess :-) We've running an Istio mesh where an ingress gateway (an Istio Envoy proxy) [routes][routing] requests to mesh services. The gateway [delegates][authz] authorisation decisions to OPA. (In the future, we're going to have one OPA instance tied to each Istio sidecar to avoid performance bottlenecks.)


[front-end.oidc]: https://github.com/StyraInc/opa-apigee-demo/blob/master/frontend-app/main.py#L72
[routing]: https://github.com/c0c0n3/kitt4sme.live/blob/main/deployment/mesh-infra/routing/ingress.yaml
[authz]: https://github.com/StyraInc/opa-apigee-demo/blob/master/frontend-app/main.py#L13

PR #331, opened 2022-04-25
proxy_init: fix version number handling

charlesdaniels:
Signed-off-by: Charles Daniels <charles@styra.com>


tsandall (approved):
LGTM. This way the version suffix on the tag gets set the VERSION variable in the Makefile. Also, the Makefile has a reasonable placeholder for local builds.

PR #335, opened 2022-05-26
New token for Alice and Bob

isacssouza:
The current tokens for Alice and Bob that are in the README have expired. I simply created new ones valid until 2031 and updated their values in the README.


anderseknert:
Thanks! See you in 2031 :)

PR #353, opened 2022-10-14
OpenTelemetry spans for gRPC ext_authz calls

vinhph0906:
I have some updated code support issues: https://github.com/open-policy-agent/opa/issues/4290. For more detail, I have reused the [distributed_tracing](https://www.openpolicyagent.org/docs/latest/configuration/#distributed-tracing) option of `opa`. But I have to export `TracerProvider` and `DistributedTracingOpts` through `plugin.manager`.  Here is a screenshot with trace on my jaeger:
image
P/s: This PR is just for review code.


internal/internal.go
```diff
@@ -117,6 +119,8 @@ func New(m *plugins.Manager, cfg *Config) plugins.Plugin {
 		server: grpc.NewServer(
 			grpc.MaxRecvMsgSize(cfg.GRPCMaxRecvMsgSize),
 			grpc.MaxSendMsgSize(cfg.GRPCMaxSendMsgSize),
+			grpc.UnaryInterceptor(otelgrpc.UnaryServerInterceptor()),
```
srenatus, line 122: Will these automatically pick up the right configuration? 🤔

srenatus (commented):
Thanks for working on this! One question inline so far.

internal/internal.go
vinhph0906: Because this line https://github.com/vinhph0906/opa-envoy-plugin/blob/%234290/vendor/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc/grpctrace.go#L51 will get `TracerProvider` from `go.opentelemetry.io/otel/internal/global`. Then we can set global `TracerProvider` at https://github.com/vinhph0906/opa-envoy-plugin/blob/%234290/vendor/github.com/open-policy-agent/opa/internal/distributedtracing/distributedtracing.go#L119. It will get the right configuration.

srenatus: OK thanks for reminding me 😅 You're right! 👍

emaincourt:
@srenatus any chance to see it merged soon ? I would love to give it a try.

sidharthramesh:
What happened to this? Such a great feature to add!

sidharthramesh:
Should the changes be based on this PR? Or have things changed too much?

PR #366, opened 2022-10-26
Upgrade base image init proxy

Minh2003:
Upgraded base image init proxy

Signed-off-by: minhlhd <minhlhd@vng.com.vn>


srenatus (approved):
👏 Thanks for contributing! 😃

PR #376, opened 2022-12-07
build: bump golang 1.19.3 -> 1.19.4

srenatus:
This is done because of https://groups.google.com/g/golang-announce/c/L_3rmdT0BMU

NOTE HOWEVER that the grpc stack doesn't use the stdlib's net/http for its HTTP2 stack, so I'm unsure about the status of that. Let's see if an advisory is published for grpc-go soon.


srenatus:
close and reopend because the checks didn't run before 🤷

PR #383, opened 2023-01-11
Fix AdmissionReview apiVersion to use v1.

naohiroo:
### What this PR does / why we need it:

I would like to propose that the apiVersion for AdmissionReview  be `admission.k8s.io/v1` instead of `admission.k8s.io/v1beta1`.

### FYI

`admission.k8s.io/v1beta1` has been removed in kubernetes docs. ( [PR](https://github.com/kubernetes/website/pull/34740/files#diff-8d3572c3fbe3cbb614db00cf1b0b090f4d7145484c9da027a1498478e0315393L409) )

### DIFF
--- a/examples/istio/quick_start.yaml
+++ b/examples/istio/quick_start.yaml
@@ -162,11 +162,14 @@ data:
   inject.rego: |
     package istio
 
+    uid := input.request.uid
+
     inject = {
-      "apiVersion": "admission.k8s.io/v1beta1",
+      "apiVersion": "admission.k8s.io/v1",
       "kind": "AdmissionReview",
       "response": {
         "allowed": true,
+        "uid": uid,
         "patchType": "JSONPatch",
         "patch": base64.encode(json.marshal(patch)),
       },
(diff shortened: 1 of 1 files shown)


examples/istio/quick_start.yaml
```diff
@@ -162,11 +162,14 @@ data:
...
 
+    uid := input.request.uid
+
     inject = {
-      "apiVersion": "admission.k8s.io/v1beta1",
+      "apiVersion": "admission.k8s.io/v1",
       "kind": "AdmissionReview",
       "response": {
         "allowed": true,
+        "uid": uid,
```
naohiroo, line 172: Needed to include the uid for the same reason as the issue below.
https://github.com/open-policy-agent/opa/issues/2467

ashutosh-narkar:
Can you please go through the steps in [this](https://www.openpolicyagent.org/docs/latest/envoy-tutorial-istio/) tutorial and ensure it works with your changes. Thanks.

ashutosh-narkar:
@naohiroo did you get a chance to verify if your changes work as expected by going through the [Istio](https://www.openpolicyagent.org/docs/latest/envoy-tutorial-istio/) tutorial?

naohiroo:
Sorry for the delay. 
I confirmed it actually worked through tutorial.

Thanks for the review.

PR #401, opened 2023-03-10
deps(build): Bump OPA v0.49.2 -> v0.50.0

ashutosh-narkar:


ashutosh-narkar:
The `Istio End-to-End Test` seems to timeout deploying the Bookinfo app which is un-related to the changes in this PR.

PR #414, opened 2023-04-28
Add grpc metrics

antgubarev:
fix the issue https://github.com/open-policy-agent/opa/issues/3852


internal/internal.go
```diff
@@ -416,14 +427,20 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 		}
 	}
 
+	totalDesicionTime := time.Since(start)
```
srenatus, line 430: typo
```suggestion
	totalDecisionTime := time.Since(start)
```

internal/internal.go
```diff
@@ -237,6 +240,15 @@ func (p *envoyExtAuthzGrpcServer) listen() {
 		addr = "grpc://" + addr
 	}
 
+	if p.cfg.EnablePrometheusMetrics {
+		summaryAuthzDuration := prometheus.NewSummary(prometheus.SummaryOpts{
+			Name: "grpc_authz_request_duration_ms",
```
srenatus, line 245: It seems like this would be the gRPC equivalent of `http_request_duration_seconds`? I guess it's unfortunate that the HTTP metrics is _"seconds"_, so "ms" might be a better choice... :+1:

srenatus (commented):
Thanks for your contribution.

The code changes looks good to me -- but adding just any sort of test would be good.

internal/internal.go
antgubarev: Sure. Most grpc calls have duration about 2-10ms.

antgubarev:
> Thanks for your contribution.
> 
> The code changes looks good to me -- but adding just any sort of test would be good.

Done

internal/internal_test.go
```diff
@@ -1596,11 +1584,27 @@ func testAuthzServerWithTruncatedBody(customLogger plugins.Plugin, dryRun, skipR
...
+	}
+
+	server := testAuthzServer(&testPlugin{}, &Config{EnablePrometheusMetrics: true})
+	output, err := server.Check(context.Background(), &req)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if output.Status.Code != int32(code.Code_OK) {
+		t.Fatal("Expected request to be allowed but got:", output)
+	}
```
srenatus, line 1603: I don't know if the test scaffolding is capable enough -- but could we check that the desired prometheus metric is actually exported, and has a non-zero value after this incoming request?

internal/internal.go
ashutosh-narkar: `grpc_authz_request_duration_ms` sgtm

internal/internal_test.go
antgubarev: @srenatus done

antgubarev: Done

antgubarev:
@srenatus @ashutosh-narkar Could you please review PR

internal/internal.go
```diff
@@ -237,6 +240,15 @@ func (p *envoyExtAuthzGrpcServer) listen() {
 		addr = "grpc://" + addr
 	}
 
+	if p.cfg.EnablePrometheusMetrics {
+		summaryAuthzDuration := prometheus.NewSummary(prometheus.SummaryOpts{
+			Name: "grpc_authz_request_duration_seconds",
```
ashutosh-narkar, line 245: Can you please help me understand the reason to use `NewSummary` v/s something like `SummaryVec` or `HistogramVec`? The latter would be useful if we want aggregates afaik. Similar to [this](https://sourcegraph.com/github.com/open-policy-agent/opa@d718975b5de4b05cf2c88d5a52efaa33e71b0f0e/-/blob/internal/prometheus/prometheus.go?L43). 

Nit: Can we change `grpc_authz_request_duration_seconds` to `grpc_request_duration_seconds`

ashutosh-narkar (commented):
Thanks for working on this @antgubarev. Few comments inline. Also these metrics should be part of the status update and decision log. It would be great if we have a test to verify that.

internal/internal.go
antgubarev: NewSummary is more simple. It's a single reason. I will change to SummaryVec if you want.

internal/internal.go
```diff
@@ -145,17 +146,18 @@ func New(m *plugins.Manager, cfg *Config) plugins.Plugin {
...
+	Path                    string `json:"path"`
+	DryRun                  bool   `json:"dry-run"`
+	EnableReflection        bool   `json:"enable-reflection"`
+	parsedQuery             ast.Body
+	ProtoDescriptor         string `json:"proto-descriptor"`
+	protoSet                *protoregistry.Files
+	GRPCMaxRecvMsgSize      int  `json:"grpc-max-recv-msg-size"`
+	GRPCMaxSendMsgSize      int  `json:"grpc-max-send-msg-size"`
+	SkipRequestBodyParse    bool `json:"skip-request-body-parse"`
+	EnablePrometheusMetrics bool `json:"enable-prometheus-metrics"`
```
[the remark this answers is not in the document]
antgubarev: Sure. Done

internal/internal.go
ashutosh-narkar: If we have multiple instances of the plugin and we need to aggregate the results, will we be able to do this?

Also is there any impact from the change in default behavior of Summary as indicated in this comment? It would be helpful if we have more tests in any case.
 ```
// A Summary captures individual observations from an event or sample stream and
// summarizes them in a manner similar to traditional summary statistics: 1. sum
// of observations, 2. observation count, 3. rank estimations.
//
// A typical use-case is the observation of request latencies. By default, a
// Summary provides the median, the 90th and the 99th percentile of the latency
// as rank estimations. However, the default behavior will change in the
// upcoming v1.0.0 of the library. There will be no rank estimations at all by
// default. For a sane transition, it is recommended to set the desired rank
// estimations explicitly.
//
// Note that the rank estimations cannot be aggregated in a meaningful way with
// the Prometheus query language (i.e. you cannot average or add them). If you
// need aggregatable quantiles (e.g. you want the 99th percentile latency of all
// queries served across all instances of a service), consider the Histogram
// metric type. See the Prometheus documentation for more details.
//
// To create Summary instances, use NewSummary.
```

**Comment by ashutosh-narkar** [comment]
@antgubarev please let us know if you need any help with this. This would be a good one to get in.

**Comment by antgubarev** [comment]
> @antgubarev please let us know if you need any help with this. This would be a good one to get in.

Thanks! I don't need any help but I need free time ) I'm going to finish it in the next few days.

**Comment by antgubarev** [comment]
> @antgubarev please let us know if you need any help with this. This would be a good one to get in.

@ashutosh-narkar I can't find a way to test status plugin :( . Would you help me?

File: internal/internal.go (hunk 4 above)
Comment (antgubarev) [reply internal/internal.go]: Yes, you're right. I fixed it. Thanks!

**Comment by antgubarev** [comment]
@ashutosh-narkar @srenatus Could you give you opinion about this PR

**Comment by ashutosh-narkar** [comment]
Thanks for working on this @antgubarev. I'll take a look next week.

File: README.md [hunk 6]
```diff
@@ -163,10 +163,14 @@ bundles:
...
-    dry-run: false
-    enable-reflection: false
+    addr: :9191 # default `:9191`
+    path: envoy/authz/allow # default: `envoy/authz/allow`
+    dry-run: false # default: false
+    enable-reflection: false # default: false
+    grpc-max-recv-msg-size: 40194304 # default: 1024 * 1024 * 4
+    grpc-max-send-msg-size: 2147483647 # default: max Int
+    skip-request-body-parse: false # default: false
+    enable-performance-metrics: false # default: false. Adds `grpc_request_duration_seconds` prometheus histogram metric
```
ashutosh-narkar, line 173: @antgubarev here we are stating that by enabling this we'll get a metric called `grpc_request_duration_seconds` which is a histogram metric. This looks good. But below we've defined `grpc_request_duration_seconds` as a `prometheus.Summary` which does not seem right. I would have imagined something like a `prometheus.HistogramVec` similar to what [OPA](https://sourcegraph.com/github.com/open-policy-agent/opa@c954a0c1eef6218b4e7f11ec5785ddc45b5a2141/-/blob/internal/prometheus/prometheus.go?L45) does. This would be a more useful metric than a summary as users typically want to understand the latency distribution of their requests. WDYT?

antgubarev: @ashutosh-narkar We've already used `prometheus.HistogramVec` not `Summary` :)

ashutosh-narkar: https://github.com/open-policy-agent/opa-envoy-plugin/pull/414/files#diff-67213f39b84a013e7db04664cd3ad943b7a237d97b9c6e862d10806ac27def28R244 here we seem to be using `Summary`.

ashutosh-narkar (commented):
Thanks for making the changes @antgubarev. Can you please look into the test failure. Also these metrics should be part of the status update and decision log. It would be great if we have a test to verify that.

antgubarev:
> Thanks for making the changes @antgubarev. Can you please look into the test failure. Also these metrics should be part of the status update and decision log. It would be great if we have a test to verify that.

@ashutosh-narkar I'm sorry It's my fault. I accidentally reverted more commits than I needed. I fixed it!

ashutosh-narkar:
@antgubarev thanks so much for working on this. The changes look good. Can you please squash your commits and rebase? Thanks.

antgubarev:
@ashutosh-narkar Done. Thanks for your review.

ashutosh-narkar (approved):
Thanks for the contribution @antgubarev!

PR #419, opened 2023-05-18
Add support for multi-valued headers.

Pushpalanka:
Relates : https://github.com/open-policy-agent/opa/issues/5925

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -306,13 +306,24 @@ func transformToHTTPHeaderFormat(input interface{}, result *http.Header) error {
 
 	takeResponseHeaders := func(headers map[string]interface{}, targetHeaders *http.Header) error {
 		for key, value := range headers {
-			var headerVal string
-			var ok bool
-			if headerVal, ok = value.(string); !ok {
-				return fmt.Errorf("type assertion error")
+			switch values := value.(type) {
+			case string:
+				targetHeaders.Add(key, values)
+			case []string:
+				for _, v := range values {
+					targetHeaders.Add(key, v)
+				}
+			case []interface{}:
+				for _, value := range values {
+					if headerVal, ok := value.(string); ok {
+						targetHeaders.Add(key, headerVal)
+					} else {
+						return fmt.Errorf("invalid value type for header '%s'", key)
+					}
+				}
+			default:
+				return fmt.Errorf("type assertion error for header '%s'", key)
 			}
-
-			targetHeaders.Add(key, headerVal)
 		}
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


ashutosh-narkar:
These changes look good @Pushpalanka. It would be helpful to have a test similar to [this](https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/internal/internal_test.go#L1261) one which exercises this functionality.

Pushpalanka:
> These changes look good @Pushpalanka. It would be helpful to have a test similar to [this](https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/internal/internal_test.go#L1261) one which exercises this functionality.

Thanks for the direction @ashutosh-narkar. A similar test is added now.

envoyauth/response.go
```diff
@@ -306,13 +306,24 @@ func transformToHTTPHeaderFormat(input interface{}, result *http.Header) error {
...
+			case string:
+				targetHeaders.Add(key, value.(string))
+			case []string:
+				for _, v := range values {
+					targetHeaders.Add(key, v)
+				}
+			case []interface{}:
+				for _, value := range values {
+					if _, ok := value.(string); ok {
+						targetHeaders.Add(key, value.(string))
```
ashutosh-narkar, line 319: Nit: What about something like

```go
if headerVal, ok := value.(string); ok {
	targetHeaders.Add(key, headerVal)
```

ashutosh-narkar (commented):
@Pushpalanka this looks good. Few comments in-line. Can you also squash your commits and author a commit message per [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines and we can then merge this. Thanks for working on this!

ashutosh-narkar:
@Pushpalanka please let us know if you need any help with this. This would be a good change to get in.

envoyauth/response.go
Pushpalanka: 👍

Pushpalanka:
> @Pushpalanka please let us know if you need any help with this. This would be a good change to get in.

Thanks, will addressed the comments and get back within today.

Pushpalanka:
@ashutosh-narkar I have addressed the review comments. There was an unrelated test failure in Istio End-to-End test which passed at a re-try, but still there is the Go Test failure which also seems unrelated to the changes. Appreciate your help.

ashutosh-narkar:
Hey @Pushpalanka, we updated the test that resulted in the Go test failure and re-ran the checks. The Go test now fails on the `TestCheckAllowObjectDecisionMultiValuedHeaders` test case. Can you look into this? If you run `go test -v -run TestCheckAllowObjectDecisionMultiValuedHeaders --count 1000`, I think you should be able to repro the issue.

ashutosh-narkar:
@Pushpalanka you're right, the test seems to pass locally. I'm going run the test suite again to rule out any transient error. We can merge after that. Thanks for working on this.

ashutosh-narkar (approved):
LGTM

PR #432, opened 2023-06-29
build: switch base images

srenatus:
After bumping the golang version to 1.20.5, we started to see this:

    ./opa_envoy_linux_amd64: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./opa_envoy_linux_amd64)
    ./opa_envoy_linux_amd64: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./opa_envoy_linux_amd64)
    ./opa_envoy_linux_amd64: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./opa_envoy_linux_amd64)

This was due to a glibc mismatch between the golang:1.20.5 image and the used base image (gcr.io/distroless).

Switching the base image to the same set of images that OPA uses resolves it.

👣 Try it locally:

1. `make image`
2. `docker run IMAGE version`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,17 +97,17 @@ image-static:
 
 image-quick:
 	sed -e 's/GOARCH/$(GOARCH)/g' Dockerfile > .Dockerfile_$(GOARCH)
-	docker build -t $(IMAGE):$(VERSION) --build-arg BASE=gcr.io/distroless/cc -f .Dockerfile_$(GOARCH) .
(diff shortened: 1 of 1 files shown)


tonychue:
Excellent, thanks for the quick turnaround and help on this issue @srenatus

charlieegan3 (approved):

PR #450, opened 2023-09-01
internal: Fix check call error being shadowed

ashutosh-narkar:
Few places in the check call shadow the error
declaration which results in some instances the
evaluation errors not being included in decision
log events. This fix attempts to resolve this.

Fixes: open-policy-agent/opa/issues/6174

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -394,19 +394,21 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 		return nil, stop, err
 	}
 
-	inputValue, err := ast.InterfaceToValue(input)
+	var inputValue ast.Value
+	inputValue, err = ast.InterfaceToValue(input)
 	if err != nil {
 		return nil, stop, err
 	}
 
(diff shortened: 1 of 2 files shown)


internal/internal.go
```diff
@@ -394,19 +394,21 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 		return nil, stop, err
 	}
 
-	inputValue, err := ast.InterfaceToValue(input)
+	var inputValue ast.Value
+	inputValue, err = ast.InterfaceToValue(input)
```
srenatus, line 398: These changes look OK, but it took me quite a moment to understand why it matters to do it _like this_. I wonder if we'd make this more robust (against future changes, too) if we changed `stop()` to be `stop(err error)` instead of relying on the single `err` reference not being shadowed.

srenatus (approved):
LGTM. I think I'd prefer something like this to be more bluntly explicit; but the change fixes the bug. Also thank you! 🎈

internal/internal.go
ashutosh-narkar: That's a good idea. I'll look into that.

PR #455, opened 2023-09-12
issue-5230

LionOnTheChase:
added traceid and spanid to decision log.

test for checking traceid in decision log is not passing

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -44,6 +44,8 @@ import (
 	"github.com/open-policy-agent/opa/tracing"
 	"github.com/open-policy-agent/opa/util"
 
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/open-policy-agent/opa-envoy-plugin/envoyauth"
 	internal_util "github.com/open-policy-agent/opa-envoy-plugin/internal/util"
 	"github.com/open-policy-agent/opa-envoy-plugin/opa/decisionlog"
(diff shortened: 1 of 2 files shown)


internal/internal.go
```diff
@@ -519,6 +521,22 @@ func (p *envoyExtAuthzGrpcServer) log(ctx context.Context, input interface{}, re
 		info.Path = p.cfg.Path
 	}
 
+	sctx := trace.SpanFromContext(ctx).SpanContext()
+	if sctx.IsValid() {
+		info.TraceID = sctx.TraceID().String()
+		info.SpanID = sctx.SpanID().String()
+	}
+
+	p.manager.Logger().WithFields(map[string]interface{}{
```
ashutosh-narkar, line 530: We should remove this log as much of this is already logged at debug level in the `check` call. This is also resulting in the test failure.

internal/internal_test.go
```diff
@@ -165,6 +167,38 @@ func TestCheckAllow(t *testing.T) {
 	}
 }
 
+func TestCheckTraceIDInAllowedRequest(t *testing.T) {
```
ashutosh-narkar, line 170: You could use something like [this](https://github.com/open-policy-agent/opa/blob/main/test/e2e/distributedtracing/distributedtracing_test.go#L148) as a reference to setup the unit test.

ashutosh-narkar (commented):
Thanks for working on this @LionOnTheChase. Few comments inline.

ashutosh-narkar:
Closing this for now. Feel free to pick this back up and address the comments when you get a chance. Thanks for looking into this.

LionOnTheChase:
@ashutosh-narkar  , thank you for your patience. I have addressed your comments and created a new PR https://github.com/open-policy-agent/opa-envoy-plugin/pull/462 . 

Kindly review

PR #474, opened 2023-10-24
issue-6333 pass input to log even if check request times out

rudrakhp:
Resolves open-policy-agent/opa#6333

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -384,13 +384,13 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 		return nil
 	}
 
-	if ctx.Err() != nil {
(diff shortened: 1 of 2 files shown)


internal/internal.go
```diff
@@ -400,6 +395,11 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 		return nil, stop, err
 	}
 
+	if ctx.Err() != nil {
```
ashutosh-narkar, line 398: We can move this before we convert to ast.Value correct ? We only need the input map to be non-nil.

rudrakhp: @ashutosh-narkar right, moved it right after parsing input.

ashutosh-narkar (approved):

PR #477, opened 2023-11-01
Added a counter metric for errors

rudrakhp:
Resolves open-policy-agent/opa#6351

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -7,6 +7,7 @@ package internal
 import (
 	"context"
 	"fmt"
+	"github.com/open-policy-agent/opa/topdown"
 	"math"
 	"net"
 	"net/url"
@@ -181,7 +182,13 @@ func New(m *plugins.Manager, cfg *Config) plugins.Plugin {
 			},
 		}, []string{"handler"})
 		plugin.metricAuthzDuration = *histogramAuthzDuration
+		errorCounter := prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "error_counter",
+			Help: "A counter for errors",
+		}, []string{"reason"})
+		plugin.metricErrorCounter = *errorCounter
 		plugin.manager.PrometheusRegister().MustRegister(histogramAuthzDuration)
+		plugin.manager.PrometheusRegister().MustRegister(errorCounter)
 	}
 
 	m.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateNotReady})
@@ -214,6 +221,7 @@ type envoyExtAuthzGrpcServer struct {
 	interQueryBuiltinCache iCache.InterQueryCache
 	distributedTracingOpts tracing.Options
 	metricAuthzDuration    prometheus.HistogramVec
+	metricErrorCounter     prometheus.CounterVec
 }
 
 type envoyExtAuthzV2Wrapper struct {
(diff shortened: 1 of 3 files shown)


internal/internal.go
```diff
@@ -392,7 +411,7 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 	}
 
 	if ctx.Err() != nil {
-		err = errors.Wrap(ctx.Err(), "check request timed out before query execution")
+		err = errors.Wrap(ctx.Err(), CheckRequestTimeoutErrorMessage)
```
srenatus, line 414: [nit] We should introduce our own error type here, and match on that type instead of the error message, please.

rudrakhp: @srenatus Thanks for the comment! I have tried to address this with minimal changes, please do review.

internal/internal.go
```diff
@@ -345,25 +353,34 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx context.Context, req *ext_authz_v3.C
...
+		internalErr.Wrap(err)
+		return nil, func() *rpc_status.Status { return nil }, &internalErr
 	}
 
 	txn, txnClose, err := result.GetTxn(ctx, p.Store())
 	if err != nil {
 		logger.WithFields(map[string]interface{}{"err": err}).Error("Unable to start new storage transaction.")
-		return nil, func() *rpc_status.Status { return nil }, err
+		internalErr = Error{Code: StartTxnErr}
+		internalErr.Wrap(err)
```
ashutosh-narkar, line 382: Nit: Since this is repeated in many places, might be better to pull into a separate routine.

internal/internal.go
srenatus: Thanks! I'll review it tomorrow unless @ashutosh-narkar beats me to it.

internal/internal.go
```diff
@@ -374,9 +389,20 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
...
+			} else if internalErr.Code != "" {
+				p.metricErrorCounter.With(prometheus.Labels{"reason": internalErr.Code}).Inc()
+			} else {
+				p.metricErrorCounter.With(prometheus.Labels{"reason": "unknown_check_error"}).Inc()
+			}
+		}
 		logErr := p.log(ctx, input, result, err)
 		if logErr != nil {
 			_ = txnClose(ctx, logErr) // Ignore error
+			p.metricErrorCounter.With(prometheus.Labels{"reason": "unknown_log_error"}).Inc()
```
ashutosh-narkar, line 405: Nit: We could probably add a debug here.

internal/internal_test.go
```diff
@@ -524,6 +524,29 @@ func TestCheckContextTimeout(t *testing.T) {
 	if len((*event.Input).(map[string]interface{})) == 0 {
 		t.Fatalf("Expected non empty input but got %v", *event.Input)
 	}
+
+	reg := prometheus.NewPedanticRegistry()
+	if err := reg.Register(server.metricErrorCounter); err != nil {
+		panic(fmt.Errorf("registering collector failed: %w", err))
```
ashutosh-narkar, line 530: Can we use `t.Fatalf` here?

internal/internal_test.go
```diff
@@ -524,6 +524,29 @@ func TestCheckContextTimeout(t *testing.T) {
...
+	}
+
+	g := prometheus.ToTransactionalGatherer(reg)
+	fam, _, err := g.Gather()
+	if err != nil {
+		panic(err)
```
ashutosh-narkar, line 536: Same as above.

ashutosh-narkar (commented):
@rudrakhp the changes look good. There seems to be some repetition in the test cases. If we reduce that by moving out common logic, that could be helpful.

rudrakhp:
Thanks for your comments @ashutosh-narkar ! Addressed them.

ashutosh-narkar (approved):
LGTM. Thanks @rudrakhp!

PR #479, opened 2023-11-06
Add support for dynamic metadata in OPA policies

tjons:
This PR resolves https://github.com/open-policy-agent/opa/issues/6209 by adding a special `dynamic_metadata` keyword to the OPA envoy plugin API. When an evaluated policy returns an object with the `dynamic_metadata` keyword, it will be written to the DynamicMetadata field of the returned CheckResponse. This allows users to emit data from OPA policies that can be consumed elsewhere in the envoy filterchain to enable richer and more dynamic auth scenarios.

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -8,10 +8,12 @@ import (
 
 	ext_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	ext_type_v3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
+	_structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/open-policy-agent/opa-envoy-plugin/internal/util"
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/topdown/builtins"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
 // EvalResult - Captures the result from evaluating a query against an input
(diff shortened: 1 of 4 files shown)


srenatus:
The code looks good! We'll need some docs for that, though, or this fine feature might go unnoticed. Would you be up to raising a companion PR against the OPA docs? 😃

tjons:
@srenatus yes, was going to ask about that. I'm adding some additional tests too to `response_test.go`. I noticed a lot of the tests in the `internal_test.go` file are testing the same features in multiple places... is there sufficient coverage in this PR or should I test dynamic metadata in some of the other areas?

tjons:
@srenatus see https://github.com/open-policy-agent/opa/pull/6389 for docs PR. I don't think there's a lot of detail needed here but let me know if you think there should be more added.

envoyauth/response.go
```diff
@@ -280,6 +282,33 @@ func (result *EvalResult) GetResponseHTTPStatus() (int, error) {
...
+// GetDynamicMetadata returns the dynamic metadata to return if part of the decision
+func (result *EvalResult) GetDynamicMetadata() (*_structpb.Struct, error) {
+	var (
+		val interface{}
+		ok  bool
+	)
+	switch decision := result.Decision.(type) {
+	case bool:
+		if decision {
+			return nil, fmt.Errorf("dynamic metadata undefined for simple 'allow'")
```
ashutosh-narkar, line 294: What about `dynamic metadata undefined for simple 'allow'` -> `dynamic metadata undefined for boolean decision`

ashutosh-narkar (commented):
The changes look good. I had just one small comment. Also having test coverage for the boolean policy decision scenario would be good. Then please squash your commits and we can get this in. Thanks @tjons!

envoyauth/response.go
tjons: done!

tjons:
updated, squashed and pushed @ashutosh-narkar! looking forward to getting this in :)

ashutosh-narkar (approved):
LGTM. Thanks @tjons!

ashutosh-narkar:
@tjons can you please fix the commit message body? There are multiple sign-offs in there. [This](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) is a good reference to author commit messages.

tjons:
@ashutosh-narkar done!

PR #481, opened 2023-11-12
Fix issue#6394: opa:0.58.0-envoy-rootless build without gid 1000

sabinahofmann:
Current behavior: Opa envoy rootless image was build without gid 1000.

Due to the request https://github.com/open-policy-agent/opa/issues/6394, the gid is now also set during the build.


ashutosh-narkar (approved):
LGTM. Thanks @sabinahofmann!

PR #491, opened 2023-11-21
Emit error metric only if prometheus metrics are enabled

rudrakhp:
Resolves: open-policy-agent/opa#6410
* Added a test case to check error handling if metrics are disabled, seg faults without this fix.
* Also removed handling empty InternalErr code as all internal errors are supposed to be created in this package and are guaranteed to have a non-empty code if there is a check error.


ashutosh-narkar (approved):
Thanks @rudrakhp!

PR #492, opened 2023-11-23
Update Istio Envoy docs to use AuthorizationPolicy instead of EnvoyFilter

tjons:
Resolves https://github.com/open-policy-agent/opa/issues/5911 by rewriting the `quick_start.yaml` file to use the Istio `AuthorizationPolicy` with a `CUSTOM` authorizer instead of the unstable `EnvoyFilter` API. I've also added some instructions to the YAML file as a comment on how to register the OPA sidecars as an external authorizer in the meshconfig object. 

~I'm working on resolving the issues with the Istio e2e tests - the mesh config object needs to be modified directly in this case to register OPA as the external authorizer so I'm figuring out how to automate that in the end-to-end tests~

### DIFF
--- a/examples/istio/quick_start.yaml
+++ b/examples/istio/quick_start.yaml
@@ -1,43 +1,53 @@
 ############################################################
-# Envoy External Authorization filter that will query OPA.
+# Add the following to the mesh config to enable external authorization:
+# mesh: |-
+#   # ADD THIS HERE
+#   extensionProviders:
+#   - name: opa-ext-authz-grpc
+#     envoyExtAuthzGrpc:
+#       service: opa-ext-authz-grpc.local
+#       port: "9191"
+#   # END
+#   defaultConfig:
+#     discoveryAddress: istiod.istio-system.svc:15012
 ############################################################
-apiVersion: networking.istio.io/v1alpha3
-kind: EnvoyFilter
+############################################################
+# AuthorizationPolicy to tell Istio to use OPA as the Authz Server
+############################################################
+apiVersion: security.istio.io/v1
+kind: AuthorizationPolicy
 metadata:
   name: ext-authz
-  namespace: istio-system
 spec:
-  configPatches:
-    - applyTo: HTTP_FILTER
-      match:
-        context: SIDECAR_INBOUND
-        listener:
-          filterChain:
-            filter:
-              name: "envoy.filters.network.http_connection_manager"
-              subFilter:
-                name: "envoy.filters.http.router"
-      patch:
-        operation: INSERT_BEFORE
-        value:
-          name: envoy.ext_authz
-          typed_config:
-            '@type': type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz
-            transport_api_version: V3
-            status_on_error:
-              code: ServiceUnavailable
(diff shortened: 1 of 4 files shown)


examples/istio/quick_start.yaml
```diff
@@ -1,43 +1,52 @@
...
-                stat_prefix: "ext_authz"
+  action: CUSTOM
+  provider:
+    # The provider name must match the extension provider defined in the mesh config.
+    # You can also replace this with sample-ext-authz-http to test the other external authorizer definition.
+    name: opa-ext-authz-grpc
+  rules:
+  - to:
+    - operation:
+        paths: ["/*"]
```
howardjohn, line 30: Do we need this? if so, why?

examples/istio/quick_start.yaml
```diff
@@ -1,43 +1,52 @@
...
+        notPaths: ["/health"]
+---
+############################################################
+# ServiceEntry to register the OPA-Istio sidecars as external authorizers.
+############################################################
+apiVersion: networking.istio.io/v1alpha3
```
howardjohn, line 36:
```suggestion
apiVersion: networking.istio.io/v1beta1
```
nit, both will work

examples/istio/quick_start.yaml
```diff
@@ -1,43 +1,52 @@
...
+kind: ServiceEntry
```
howardjohn, line 37: nit: mind be nice to add `exportTo: [.]` so it doesn't impact other namespaces

test/bats/istio-cm-patch.yaml
```diff
@@ -0,0 +1,33 @@
+data:
```
howardjohn, line 1: Not sure the testing setup, but this seems to patch a lot more than what is needed. Maybe thats intended, though

howardjohn (commented):
LGTM (as an Istio maintainer)

test/bats/istio-cm-patch.yaml
tjons: this is intended. I found it impossible to just add the `extensionProvider` to the mesh config since is a nested yaml object... it doesn't play nicely with `kubectl patch` even with a json patch. Any thoughts? Otherwise this is just to restore the rest of the default data that is created when we install istio with the demo profile for our CI

examples/istio/quick_start.yaml
tjons: 👍

examples/istio/quick_start.yaml
tjons: will fix

examples/istio/quick_start.yaml
tjons: nope! unnecessary

tjons:
@howardjohn addressed your comments - LMK if anything else should be addressed

ashutosh-narkar (commented):
LGTM. You may have already done this but have you tried these steps manually to ensure everything works end to end? 

If you could squash your commits we can get this in.

tjons:
@ashutosh-narkar yes, have tested manually a bunch of different ways. Squashed and pushed

ashutosh-narkar (approved):
Thanks @tjons!

PR #503, opened 2024-01-11
Decision Id in check response

PietervdWerk:
Fixes https://github.com/open-policy-agent/opa/issues/6519 by adding decision id to the dynamic metadata.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -557,6 +557,19 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 		}
 	}
 
+	// Add decision_id to dynamic metadata
+	if resp.DynamicMetadata == nil {
+		resp.DynamicMetadata = &_structpb.Struct{
+			Fields: map[string]*_structpb.Value{},
+		}
+	}
+
+	resp.DynamicMetadata.Fields["decision_id"] = &_structpb.Value{
+		Kind: &_structpb.Value_StringValue{
+			StringValue: result.DecisionID,
+		},
+	}
+
 	return resp, stop, nil
 }
 
--- a/internal/internal_test.go
+++ b/internal/internal_test.go
@@ -317,6 +317,28 @@ func TestCheckDeny(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


internal/internal.go
```diff
@@ -557,6 +557,18 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
...
+			Fields: map[string]*_structpb.Value{},
+		}
+	}
+
+	resp.DynamicMetadata.Fields["decision_id"] = &_structpb.Value{
+		Kind: &_structpb.Value_StringValue{
+			StringValue: result.DecisionID,
+		},
+	}
+
```
ashutosh-narkar, line 571: What if we first check if the decision id is defined and if it is then we include it in the dynamic metadata? Currently we'll include dynamic metadata always and it could possibly not be that useful in some cases.

ashutosh-narkar (commented):
Hey @PietervdWerk thanks for picking this up. If you could add some unit tests that would be great.

internal/internal.go
PietervdWerk: The decision id is always a uuid4, if it errors the error is returned by the check function so it looks to me like that would be a redundant check.

ashutosh-narkar:
@PietervdWerk the changes look good. Can you please sign-off your commit and we can get this in. Thanks.

srenatus (approved):
Thanks!! 👏

ashutosh-narkar:
@PietervdWerk something may gotten missed up when you rebased. The `golang.org/x/tools` update should not be part of your change. Can you please fix that and push a single commit with your change. Also thanks for fixing the sign-off.

PietervdWerk:
> @PietervdWerk something may gotten missed up when you rebased. The `golang.org/x/tools` update should not be part of your change. Can you please fix that and push a single commit with your change. Also thanks for fixing the sign-off.

I've cleaned the main history and rebased my commit on it. Sorry about that, first time signing off on a commit. Should be all good now :).

ashutosh-narkar (approved):
LGTM. Thanks @PietervdWerk!

PR #512, opened 2024-02-01
issue-4687 Add B3 and Jaeger context propagators

be-a-bee:


be-a-bee:
@ashutosh-narkar , i'm adding a new module dependency and hence i'm getting the following error

internal/internal.go:29:2: cannot find module providing package go.opentelemetry.io/contrib/propagators/b3: import lookup disabled by -mod=vendor

how do I fix this ? should I merge any additional files along with this ? 

Sorry if it is a trivial question. just that i'm not experienced enough with golang projects.

ashutosh-narkar:
I'd imagine if you do a `go mod tidy` and `go mod vendor`, the necessary dependency should be pulled in.

be-a-bee:
@ashutosh-narkar  , thank you that was helpful. 

can you please review this PR

ashutosh-narkar (commented):
@be-a-bee thanks for the contribution. It would be good if we can avoid the version update in this PR. If we can limit this change to only adding the B3 propagators that would be ideal. Thanks.

be-a-bee:
Sure. will create a separate PR with the relevant changes alone.

PR #516, opened 2024-02-08
Allow to set the decisionID from an external function

Pushpalanka:
## Short description 
An improvement to allow sending the decision ID from outside avoiding the default UUID  generation [here](https://github.com/open-policy-agent/opa-envoy-plugin/blob/543adde23777f285a2e974b18d509642b3c3bf8d/envoyauth/response.go#L44)

## Expected behavior
Suggesting improvement will allow passing a function from outside which can operate on the result. Current behavior of the default UUID generation is left intact, unless the function set a decision ID.

## Additional context
In a benchmark test we noticed the UUID generation is comparatively costly. (Below image shows a branch of it.)

Screenshot 2024-02-08 at 17 13 34

Hence it will benefit if we can send an externally generated ID and set it here, avoiding the costly operation.

CC: @mjungsbluth

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -35,13 +35,20 @@ type StopFunc = func()
 type TransactionCloser func(ctx context.Context, err error) error
 
 // NewEvalResult creates a new EvalResult and a StopFunc that is used to stop the timer for metrics
-func NewEvalResult() (*EvalResult, StopFunc, error) {
+func NewEvalResult(opts ...func(*EvalResult)) (*EvalResult, StopFunc, error) {
 	var err error
 
-	er := EvalResult{
+	er := &EvalResult{
 		Metrics: metrics.New(),
 	}
-	er.DecisionID, err = util.UUID4()
+
+	for _, opt := range opts {
+		opt(er)
+	}
+
+	if er.DecisionID == "" {
+		er.DecisionID, err = util.UUID4()
+	}
 
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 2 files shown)


envoyauth/response.go
```diff
@@ -34,14 +34,24 @@ type StopFunc = func()
 // TransactionCloser should be called to abort the transaction
 type TransactionCloser func(ctx context.Context, err error) error
 
+// An Opt allowing externally setting the decision ID
+type Opt func(*EvalResult)
```
ashutosh-narkar, line 38: Nit: do we need to define this?

envoyauth/response.go
```diff
@@ -34,14 +34,24 @@ type StopFunc = func()
...
+
 // NewEvalResult creates a new EvalResult and a StopFunc that is used to stop the timer for metrics
-func NewEvalResult() (*EvalResult, StopFunc, error) {
+func NewEvalResult(opts ...Opt) (*EvalResult, StopFunc, error) {
```
ashutosh-narkar, line 41: Can we do something like:

```go
func NewEvalResult(opts ...func(*EvalResult)) (*EvalResult, StopFunc, error)
```

ashutosh-narkar (commented):
Thanks for working on this @Pushpalanka! The changes lgtm. Just few comments below and we can then get this in.

envoyauth/response.go
Pushpalanka: We can remove it. Thanks for the suggestion.

envoyauth/response.go
Pushpalanka: Yes, improved.

ashutosh-narkar (approved):
Thanks @Pushpalanka!

Pushpalanka:
> Thanks @Pushpalanka!

Thanks for your support as well. :)

PR #518, opened 2024-02-12
add b3singleheader injection

be-a-bee:
add b3singleheader injection
add test to assert propagation of b3 single and multi headers.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -135,11 +135,11 @@ func New(m *plugins.Manager, cfg *Config) plugins.Plugin {
 	if m.TracerProvider() != nil {
 		grpcTracingOption := []otelgrpc.Option{
 			otelgrpc.WithTracerProvider(m.TracerProvider()),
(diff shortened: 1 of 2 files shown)


be-a-bee:
@ashutosh-narkar , please review.

ashutosh-narkar (approved):
@be-a-bee thanks for looking into this.

be-a-bee:
@ashutosh-narkar  , when can we expect the next release ?

ashutosh-narkar:
We have a monthly release cycle. There will be a release end of the month.

be-a-bee:
Thanks @ashutosh-narkar  . is there a way we can have a nightly build kind of thing so that we can consume these changes in our environment ? Or any other alternative ? Please advise.

ashutosh-narkar:
You can use for example the `openpolicyagent/opa:0.61.0-envoy-3` image which should have your changes.  See this [job](https://github.com/open-policy-agent/opa-envoy-plugin/actions/runs/7877191231/job/21492848173) for details.

PR #547, opened 2024-05-02
Specific response assertion error messages

PietervdWerk:
I've been getting the error `"failed to get response body: type assertion error"` and it took me quite a while to figure out what the problem was. I've copied my policy into a playground and it worked fine there (in hindsight that's obvious).

I propose a more verbose and specific error message when assertion fails. I've taken the liberty to come with some suggestions.

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -105,7 +105,7 @@ func (result *EvalResult) IsAllowed() (bool, error) {
 		}
 
 		if allowed, ok = val.(bool); !ok {
-			return false, fmt.Errorf("type assertion error")
+			return false, fmt.Errorf("type assertion error, expected allowed to be of type 'boolean' but got '%T'", val)
 		}
 
 		return allowed, nil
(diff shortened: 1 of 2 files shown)


envoyauth/response.go
```diff
@@ -307,7 +307,7 @@ func (result *EvalResult) GetDynamicMetadata() (*_structpb.Struct, error) {
 
 		metadata, ok := val.(map[string]interface{})
 		if !ok {
-			return nil, fmt.Errorf("type assertion error")
+			return nil, fmt.Errorf("type assertion error, dynamic_metadata is not a map[string]interface{}")
```
anderseknert, line 310: `map[string]interface{}` is a Go type, not a Rego one. I'd just say "is not an object" here. Same for headers below.

anderseknert (commented):
Thanks! Looks good to me, with a minor change suggested. I wonder if it would be helpful to go one step further and say "expected x, got y" in these messages? But this is an improvement nonetheless.

anderseknert (dismissed):
LGTM! @ashutosh-narkar ?

ashutosh-narkar (commented):
@PietervdWerk the changes lgtm. It would be great if we can update the tests to reflect these detailed error messages.

ashutosh-narkar:
@PietervdWerk this would be a good change to get in. Are you able to update the tests?

PietervdWerk:
Yes tomorrow! (I had a few days off, back to work tomorrow.)

ashutosh-narkar (approved):
Thanks @PietervdWerk!

PR #558, opened 2024-06-11
Remove rootless images from build process 

tjons:
Resolves https://github.com/open-policy-agent/opa/issues/6810 by removing rootless images from the `Makefile`, and thus, from the build/release process for `opa-envoy-plugin`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,13 +8,11 @@ FROM ${BASE}
 
 # Any non-zero number will do, and unfortunately a named user will not, as k8s
 # pod securityContext runAsNonRoot can't resolve the user ID:
-# https://github.com/kubernetes/kubernetes/issues/40958. Make root (uid 0) when
-# not specified.
-ARG USER=0
+# https://github.com/kubernetes/kubernetes/issues/40958.
+ARG USER=1000:1000
+USER ${USER}
 
 MAINTAINER Ashutosh Narkar  <anarkar4387@gmail.com>
-
-USER ${USER}
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


charlieegan3:
Hey Tyler, thanks for picking this one up!

I think the first thing we need to do is to set the user arg to non root: https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/Dockerfile#L13C1-L13C11, similar to this: https://github.com/open-policy-agent/opa/blob/b463d30028d927a00bf0143d6fbf0533bf5d6783/Dockerfile#L15

Then we can get a release out and share that that the default user has been updated to non-root.


Either in the same release, or in a follow on release, we can do as you have here, where we drop the rootless tags. We've made a big noise about the end of rootless images in the past for opa, but it might be  good to do a final reminder in the next release for opa-envoy too.

@ashutosh-narkar might be able to provide some metrics on the image pulls for the rootless tags. Also Ash, let me know if you think that plan above sounds ok, and if you think we can do the user change and the dropping of the tags in the same release.

charlieegan3:
```
$ docker image inspect openpolicyagent/opa:latest-envoy-static | jq .[0].Config.User
"0"
```

😞

tjons:
@charlieegan3 thanks for the pointer about the default user, I missed that part! Would you like me to restore the `rootless` build targets for that last release? If there are no users I'm not sure that it's necessary but I like the thought you put into this.

charlieegan3:
Hey, I think that the best course of action is to drop rootless tag at the same time as making the default image use a non-root user.

The rationale here being that we've already spent a lot of time letting users know that rootless images are discontinued and that they are not to be used. Making this change now in opa-envoy will be 'just' updating opa-envoy to be the same as OPA.

Do you think that makes sense too?

charlieegan3 (approved):
Thanks for the work here @tjons 👏

charlieegan3:
When merging, we should rebase and merge https://github.com/open-policy-agent/opa-envoy-plugin/pull/560 too.

anderseknert (approved):
LGTM!

PR #573, opened 2024-07-31
Issue 6901: set invalid request status if request parsing fails

rudrakhp:
Resolves open-policy-agent/opa#6901

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -370,22 +370,22 @@ func (p *envoyExtAuthzGrpcServer) Check(ctx context.Context, req *ext_authz_v3.C
 func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*ext_authz_v3.CheckResponse, func() *rpc_status.Status, *Error) {
 	var err error
 	var evalErr error
-	var internalErr Error
+	var internalErr *Error
 	start := time.Now()
 	logger := p.manager.Logger()
 
 	result, stopeval, err := envoyauth.NewEvalResult()
 	if err != nil {
 		logger.WithFields(map[string]interface{}{"err": err}).Error("Unable to start new evaluation.")
-		internalErr = internalError(StartCheckErr, err)
-		return nil, func() *rpc_status.Status { return nil }, &internalErr
+		internalErr = newInternalError(StartCheckErr, err)
+		return nil, func() *rpc_status.Status { return nil }, internalErr
 	}
 
 	txn, txnClose, err := result.GetTxn(ctx, p.Store())
 	if err != nil {
 		logger.WithFields(map[string]interface{}{"err": err}).Error("Unable to start new storage transaction.")
-		internalErr = internalError(StartTxnErr, err)
-		return nil, func() *rpc_status.Status { return nil }, &internalErr
+		internalErr = newInternalError(StartTxnErr, err)
+		return nil, func() *rpc_status.Status { return nil }, internalErr
 	}
 
 	result.Txn = txn
@@ -398,9 +398,9 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
 		stopeval()
 		if p.cfg.EnablePerformanceMetrics {
 			var topdownError *topdown.Error
-			if internalErr.Unwrap() != nil && errors.As(internalErr.Unwrap(), &topdownError) {
+			if internalErr != nil && errors.As(internalErr.Unwrap(), &topdownError) {
 				p.metricErrorCounter.With(prometheus.Labels{"reason": topdownError.Code}).Inc()
-			} else if internalErr.Code != "" {
+			} else if internalErr != nil && internalErr.Code != "" {
 				p.metricErrorCounter.With(prometheus.Labels{"reason": internalErr.Code}).Inc()
 			}
 		}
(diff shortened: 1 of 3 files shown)


internal/internal.go
```diff
@@ -422,27 +422,41 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req interface{}) (*
...
+			HttpResponse: &ext_authz_v3.CheckResponse_DeniedResponse{
+				DeniedResponse: &ext_authz_v3.DeniedHttpResponse{
+					Status: &ext_type_v3.HttpStatus{
+						Code: ext_type_v3.StatusCode(ext_type_v3.StatusCode_BadRequest),
+					},
+					Body: internalErr.Error(),
+				},
+			},
+			DynamicMetadata: nil,
+		}, stop, internalErr
```
ashutosh-narkar, line 440: I think we need to verify here if Envoy returns `403` by default or `400`. It's possible if you return an error along with the response, `403` will be the default. Would be good to verify this.

ashutosh-narkar: > Looks like setting the DeniedHttpResponse.Status field allows the authz service to send an HTTP response to the downstream client.

Sure but what happens when the authz service returns both an error and a response. Have you verified the status from the response is picked up?

ashutosh-narkar: > Code pointers:
[v3/external_auth.pb.go#L280-L283](https://github.com/envoyproxy/go-control-plane/blob/main/envoy/service/auth/v3/external_auth.pb.go#L280-L283)
[v3/external_auth.pb.go#L83-L85](https://github.com/envoyproxy/go-control-plane/blob/main/envoy/service/auth/v3/external_auth.pb.go#L83-L85)

Looks like the response status will be picked up but good to verify in any case.

rudrakhp: On a second glance looks like my `CheckResponse.Status` must be `403` but `DeniedHttpResponse.Status` can be `401`. That case Envoy treats it as a denied request but sends `401` HTTP status code downstream. WDYT @ashutosh-narkar ?

ashutosh-narkar: I think `CheckResponse.Status` can be any non-200 and it will considered as deny. `DeniedHttpResponse.Status` overrides it and is sent downstream. So if you just set `CheckResponse.Status` to say `400` will Envoy still send `403` to the client? If no, then you can probably avoid `DeniedHttpResponse`.

rudrakhp: I modified the PR, the `CheckResponse.Status` is `200 OK` but downstream response in `403 Forbidden`. Do checkout the links above and let me know WYT.

rudrakhp: @ashutosh-narkar we can't set `CheckResponse.Status` as non-200, this is because for all non-OK responses `status_on_error` sends default `403`. Check [this](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto#L96) out.

ashutosh-narkar: > @ashutosh-narkar we can't set CheckResponse.Status as non-200, this is because for all non-OK responses status_on_error sends default 403.

But you can override it in the denied response. I think the way you had it originally, ie `403` for `CheckResponse.Status` and `400` for `DeniedHttpResponse.Status` seemed correct to me.

rudrakhp: @ashutosh-narkar I went through envoy control plane docs. I think the right thing to do would be as suggested by you, return a `403` CheckResponse Status and a `401` HTTP downstream status. The error should be nil though. Updated the PR accordingly. Please review. Thanks!

ashutosh-narkar (dismissed):

ashutosh-narkar:
@rudrakhp please squash your commits and we can get this in.  Also have you tested this end-to-end?

rudrakhp:
> @rudrakhp please squash your commits and we can get this in. Also have you tested this end-to-end?

I added a invalid URL case in the e2e test suite, but somehow it's getting a 404 before 400 bad request. @ashutosh-narkar if you are aware of this do let me know, otherwise will come back to this tomorrow.

ashutosh-narkar:
> I added a case for invalid URL in the e2e test suite, but somehow it's getting a 404 before 400 bad request. @ashutosh-narkar if you are aware of this do let me know, otherwise will come back to this tomorrow.

Is the request even reaching OPA?

ashutosh-narkar:
@rudrakhp did you get a chance to look into the test failures. This would be a good one to get in.

rudrakhp:
> @rudrakhp did you get a chance to look into the test failures. This would be a good one to get in.

@ashutosh-narkar was out on vacation, will get something out in the next couple of days.

rudrakhp:
@ashutosh-narkar looks like the place where I was adding the test was for testing istio setup with the plugin enabled but requests were not going via OPA. I don't see any E2E suite where I can add this test, let me know if there is any. I think unit tests should suffice to test this behaviour as we rely on go-control-plane anyways. WDYT?

ashutosh-narkar:
@rudrakhp this looks good. If you could just update the commit message with a note explaining the change that would be great. We can then get this in. Thanks for the contribution.

rudrakhp:
@ashutosh-narkar updated commit message

ashutosh-narkar (approved):

PR #575, opened 2024-08-02
Upgrade CI dependencies for testing

tjons:
Istio 1.19.4 (the current test target) is EOL at this point. This PR upgrades the test version for Istio CI tests to 1.22.3, which is latest. See https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases for more.  In the process, I also upgraded `kind` to latest, as the `0.11` kubernetes version (1.21) is no longer supported by the Istio project.

While updating this, I noticed that our test version of Envoy was 1.26, which is also unsupported: https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history, so I fixed that as well.

I don't believe we need to bump these versions after every Istio or Envoy patch/minor release, but it would be nice to make sure that we're still testing against maintained versions.

### DIFF
--- a/build/install-istio-with-kind.sh
+++ b/build/install-istio-with-kind.sh
@@ -6,8 +6,8 @@ set -x
 
 GOARCH=$(go env GOARCH)
 GOOS=$(go env GOOS)
-KIND_VERSION=0.11.1
-ISTIO_VERSION=1.19.4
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (approved):
Thanks for the contribution @tjons!

PR #603, opened 2024-10-11
Update GO version to match OPA version

pratimsc:
The OPA Envoy Plugin fails due to CVEs present in G 1.22.5. Update it to Go version used by OPA i.e. 1.23.1, which has the required CVE fixes.

List of CVEs:
- CVE-2024-34155
- CVE-2024-34156
- CVE-2024-34158


tjons:
@pratimsc I think it would be a good idea to bump the `go.mod` version here, can you add that?

pratimsc:
@tjons  - I have updated the version in go.mod file

nfuden:
Does it make sense to add a ci check at some point to make sure the toolchain / mod version are the same as in .go-version?
After all go-version is used by the makefile. 
Definitely out of scope for this just a (potentially) interesting thought

pratimsc:
> Does it make sense to add a ci check at some point to make sure the toolchain / mod version are the same as in .go-version? After all go-version is used by the makefile. Definitely out of scope for this just a (potentially) interesting thought

@nfuden  - It will be good idea to put in the check, and also another check to use the same version of Go as the `opa` itself. So we have consistency across the `opa` and `envoy-plugin`. 

Is the present PR good to merge?

tjons:
@pratimsc neither @nfuden nor I are maintainers, I'm just a contributor. cc @ashutosh-narkar to review.

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/open-policy-agent/opa-envoy-plugin
 
-go 1.21
-toolchain go1.22.5
```
ashutosh-narkar, line 4: Thanks for updating the Go version. The min Go version to build this module has changed. Why is that needed?

pratimsc: @ashutosh-narkar - As mentioned in the PR's description, this to ensure that it aligns with version using by main OPA binary (Ref: https://github.com/open-policy-agent/opa/blob/main/.go-version)

And also to mitigate the CVEs listed in the PR.

ashutosh-narkar: I get that we should update the Go version. I'm asking about the minimum version to build the module has also updated and wanted to know why that is? If you check OPA's [go.mod](https://github.com/open-policy-agent/opa/blob/main/go.mod#L3) file the minimum version was not updated.

pratimsc: @ashutosh-narkar  - removed the update of minimum go version.

ashutosh-narkar (approved):
LGTM

PR #604, opened 2024-10-14
Refactor creation of prepared queries

mjungsbluth:
Before this refactoring only the PreparedQuery options could be passed when using this plugin as a library. There are two more options (prepare options and evaluation options) that can be passed when using the OPA library directly.

This change breaks the interface but also allows for extensibility of creating the prepared queries in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+- Refactor creation of prepared queries (#604, @mjungsbluth)
+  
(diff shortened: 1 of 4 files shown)


ashutosh-narkar (commented):
The changes look good. We should add a note in the change log about the interface change so that we can include it in the release notes. Also do the existing tests exercise the new changes?

mjungsbluth:
I will add a note to the changelog

As for tests: they cover all changes, however there never was coverage for passing options to the Eval method, I will add a minimal test.

We‘ll also integrate this in Skipper to do partial evaluation on prepare as an optimization. So we‘ll know how well the changed API works design wise…

mjungsbluth:
@ashutosh-narkar I added an item to the changelog and also added a test that verifies a user of a library can modify the existing behaviour.

ashutosh-narkar (approved):
LGTM. Thanks @mjungsbluth!

PR #610, opened 2024-10-31
Disable git safe directory check

ashutosh-narkar:


build/build-release.sh
```diff
@@ -46,6 +46,7 @@ build_release() {
 }
 
 clone_repo() {
+    git config --system --add safe.directory '*'
```
ashutosh-narkar, line 49: We could disable the check for a specify directory by moving this command in the Makefile similar to what OPA does.

PR #614, opened 2024-11-13
Adding support for envoy ext_proc in OPA

pweiber:
This PR resolves [open-policy-agent/opa#6639](https://github.com/open-policy-agent/opa/issues/6639) by adding the support for the usage of envoy external processor(ext_proc).

This PR also includes a plugin refactoring to be able to run different plugins in different ports based on the config file, for example: 
authz at :9191 and ext_proc at :9292.

Consider this PR as a first draft for the functionality so we can validate the feature first.

All suggestions are welcome and I'll try to fix the issues found as soon as possible.


plugin/plugin.go
```diff
@@ -10,18 +10,34 @@ import (
...
-// New returns the object initialized with a valid plugin configuration.
-func (Factory) New(m *plugins.Manager, config interface{}) plugins.Plugin {
-	return internal.New(m, config.(*internal.Config))
+// Plugin names to register with the OPA plugin manager.
+const (
+	AuthZPluginName   = "envoy_ext_authz_grpc"
+	ExtProcPluginName = "envoy_ext_proc_grpc"
+)
+
+// New method for AuthZFactory.
```
ashutosh-narkar, line 25: So this change would be break backward compatibility, correct? The rest of the changes seem backward compatible. Is there a way to either keep the old one and if not we can deprecate it in favor of a more flexible implementation.

pweiber: I'm not sure to be honest, if we consider the changed structure It'll start the authz plugin with "envoy_ext_authz_grpc" but the wrapper for the v2 is still there, unless it requires the naming to be "envoy.ext_authz.grpc" then yes it'll break.

If it can use the same NewAuthZ internal method what we can do is something like this:

plugin.go
```go
type Factory struct{}

func (Factory) New(m *plugins.Manager, config interface{}) plugins.Plugin {
	return internal.NewAuthZ(m, config.(*internal.Config))
}

func (Factory) Validate(m *plugins.Manager, configBytes []byte) (interface{}, error) {
	return internal.Validate(m, configBytes)
}
``` 

and for the main.go we move back to use the Factory
```go
runtime.RegisterPlugin("envoy.ext_authz.grpc", plugin.Factory{})
``` 

What do you think? I'm not sure how to test it, so if you can provide me with some guidance we can check if it works

pweiber: Seems like it worked, I'll adjust the logs and work on the tests for the new files, let me know if you have any other concerns about the backward compatibility

ashutosh-narkar: The rest seemed ok to me. Thanks for working on this!

tjons:
@pweiber I do apologize for our dropping this PR! Are you still interested in getting this in? If so, let me know and I can work with you to polish this, add docs, and ensure that it's ready for merge. If not, no worries - I'll close this if I don't hear from you by 4/12/2025

pweiber:
> @pweiber I do apologize for our dropping this PR! Are you still interested in getting this in? If so, let me know and I can work with you to polish this, add docs, and ensure that it's ready for merge. If not, no worries - I'll close this if I don't hear from you by 4/12/2025

@tjons I still believe that would be a great addition, I'll probably push some changes and unit tests tomorrow/wed and you can also take a look on that, but let's polish this feature, if you need any access to my fork let me know, I allowed edits to the mainteners so you should be good to change anything that is needed as well

envoyextproc/evaluation.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/open-policy-agent/opa/v1/storage"
+	"github.com/open-policy-agent/opa/v1/topdown/builtins"
+	iCache "github.com/open-policy-agent/opa/v1/topdown/cache"
+	"github.com/open-policy-agent/opa/v1/topdown/print"
+	"github.com/open-policy-agent/opa/v1/tracing"
+)
+
+// EvalContext - This is an SPI that has to be provided if the envoy external authorization
+// is used from outside the plugin, i.e. as a Go module
+type EvalContext interface {
```
tjons, line 21: This interface is identical to the one in `envoyauth`. Let's refactor this, and the `PrepareQueryOpts` type below into a single interface in a shared package and alias the one in `envoyauth` to refer to the shared package.

envoyextproc/request.go
```diff
@@ -0,0 +1,473 @@
...
+	fmt.Sprintf("Parsed query parameters: %v", parsedQueryInterface)
+
+	return parsedPathInterface, parsedQueryInterface, nil
+}
+
+func getParsedBody(logger logging.Logger, headers map[string]string, body string, rawBody []byte, parsedPath []interface{}, protoSet *protoregistry.Files) (interface{}, bool, error) {
+	var data interface{}
+
+	if val, ok := headers["content-type"]; ok {
+		if strings.Contains(val, "application/json") {
```
tjons, line 233: suggest pulling these content-type specific body parsers out into their own methods

examples/extproc/client.go
```diff
@@ -0,0 +1,349 @@
+package main
```
tjons, line 1: Note: we'll need to add docs to the main website once this merges

envoyextproc/response.go
```diff
@@ -0,0 +1,436 @@
...
+	"github.com/open-policy-agent/opa/v1/bundle"
+	"github.com/open-policy-agent/opa/v1/metrics"
+	"github.com/open-policy-agent/opa/v1/storage"
+	"github.com/open-policy-agent/opa/v1/topdown/builtins"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// EvalResult captures the result from evaluating a query against an input.
+type EvalResult struct {
+	Revision       string // Deprecated: Use `revisions` instead.
```
tjons, line 22: We can drop this :) we don't need to preserve deprecated fields in a new implementation!

internal/internal.go
```diff
@@ -755,3 +759,478 @@ func v2Status(s *ext_type_v3.HttpStatus) *ext_type_v2.HttpStatus {
 		Code: ext_type_v2.StatusCode(s.Code),
 	}
 }
+
+// envoyExtProcGrpcServer represents the ext_proc gRPC server implementation.
```
tjons, line 763: let's split all the `extproc` related code out into it's own subpackage. this file is already really long...

internal/internal_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
```
tjons, line 11: same note on splitting the tests out here

internal/internal_test.go
```diff
@@ -2497,3 +2501,1502 @@ func (p *testPluginError) Log(_ context.Context, event logs.EventV1) error {
...
+	if !ok {
+		return fmt.Errorf("expected ProcessingRequest")
+	}
+
+	received, err := m.Recv()
+	if err != nil {
+		return err
+	}
+
+	*req = *received
```
tjons, line 2570: this pointer deference and copy operation copies a mutex which is causing the linter to fail... do we need this mock implementation to actually copy the object?

tjons (commented):
Some comments from initial review... I'll work on some of these too, likely starting with the package refactor.

tjons:
@pweiber I'm focused on (finally!) wrapping this up this month. I can take over your branch, but could you please sign the DCO so we can use these commits?

pweiber:
> @pweiber I'm focused on (finally!) wrapping this up this month. I can take over your branch, but could you please sign the DCO so we can use these commits?

@tjons Seems like it solved the dco from my end but it messed up when I tried to solve the conflict during the rebase, for some reason it is expecting "Tyler Schade" now instead of "tjons" but it was signed as "tjons" originally. 

Sorry for the delay, I'm currently taking some time off, it should be unblocked now for you, but let me know If I need to fix that from my side, and I'll do it again.

tjons:
Thanks @pweiber for the contribution, this is a valuable first-pass. I'm going to be closing this PR since it's grown rather stale and starting with a different approach as I'm hoping to implement this before Kubecon NA in ~5 weeks. I appreciate your contribution to the project and am sorry for how this got dropped, grew stale, etc.

PR #621, opened 2024-12-13
feat: return query_parameters_to_remove with OkResponse

regeda:
Envoy External Auth v3 API supports the `query_parameters_to_remove` attribute with `OkResponse`.

See https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#service-auth-v3-okhttpresponse

The change allows to return from OPA which query parameters should be removed from the upstream query.

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -114,41 +114,49 @@ func (result *EvalResult) IsAllowed() (bool, error) {
 	return false, result.invalidDecisionErr()
 }
 
-// GetRequestHTTPHeadersToRemove - returns the http headers to remove from the original request before dispatching
-// it to the upstream
-func (result *EvalResult) GetRequestHTTPHeadersToRemove() ([]string, error) {
-	headersToRemove := []string{}
-
+func (result *EvalResult) getStringSliceFromDecision(fieldName string) ([]string, error) {
 	switch decision := result.Decision.(type) {
 	case bool:
-		return headersToRemove, nil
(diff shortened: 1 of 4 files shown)


SarkanMo-tomtom (approved):

envoyauth/response.go
```diff
@@ -114,6 +114,43 @@ func (result *EvalResult) IsAllowed() (bool, error) {
 	return false, result.invalidDecisionErr()
 }
 
+// GetRequestQueryParametersToRemove - returns the query parameters to remove from the original request before dispatching
+// it to the upstream
+func (result *EvalResult) GetRequestQueryParametersToRemove() ([]string, error) {
```
tjons, line 119: Could you add a unit test for this method? Would be nice to make sure that changes don't regress in the future.

tjons (changes requested):
Thanks for contributing! I'm excited about this feature - one request for some additional unit testing...

envoyauth/response.go
regeda: Sure! I reworked `GetRequestHTTPHeadersToRemove` too bc `GetRequestQueryParametersToRemove` does the same.

tjons:
This looks good to me... we will need to update the docs here: https://github.com/open-policy-agent/opa/blob/main/docs/content/envoy-primer.md#output-document.

tjons (approved):

PR #628, opened 2025-01-02
refact: unify get headers approach from a decision

regeda:
Both `headers` and `response_headers_to_add` attributes returned by the same function.

Remove intermediate `http.Header` transformation between decision output and Envoy headers.

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"slices"
 
 	ext_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	ext_type_v3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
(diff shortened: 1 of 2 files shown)


envoyauth/response.go
```diff
@@ -346,74 +310,64 @@ func (result *EvalResult) GetResponseEnvoyHTTPStatus() (*ext_type_v3.HttpStatus,
...
-				for _, value := range values {
-					if headerVal, ok := value.(string); ok {
-						targetHeaders.Add(key, headerVal)
-					} else {
-						return fmt.Errorf("invalid value type for header '%s'", key)
-					}
+func makeEnvoyHeaderValueOptionsFromHeadersMap(hvo []*ext_core_v3.HeaderValueOption, headers map[string]any) ([]*ext_core_v3.HeaderValueOption, error) {
+	hvo = slices.Grow(hvo, len(headers))
+	for key, value := range headers {
+		key = textproto.CanonicalMIMEHeaderKey(key)
```
srenatus, line 325: If we didn't do this before, it might be worth calling it out on the change log. It's the right thing to do, I suppose, but could potentially break someone's policy if they depended on the headers coming out like they went in.

srenatus (dismissed):

envoyauth/response.go
regeda: Previously the plugin used `http.Header` where the function `textproto.CanonicalMIMEHeaderKey` has been approached internally.

srenatus: Ah wait, I was misreading this before -- we're normalizing the header keys "on the way out", so when preparing the CheckResponse for Envoy. So, it probably isn't needed -- [Envoy will normalize the keys to all-lowercase anyways](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/header_casing), if I'm reading this right.

So the only effect here would be case-related clashes -- if a policy returns headers like `{"foo": "bar", "Foo": "baz"}`, then this line will make the output `{"Foo": "..."}` with a non-deterministic outcome. Before, we'd rely on Envoy to decide, I guess?

In the interest of caution, I'd keep the current behaviour and drop this line. What do you think?

regeda: I've removed normalization of the key.

srenatus (approved):

envoyauth/response.go
```diff
@@ -159,68 +160,30 @@ func (result *EvalResult) GetRequestHTTPHeadersToRemove() ([]string, error) {
 	return result.getStringSliceFromDecision("request_headers_to_remove")
 }
 
-// GetResponseHTTPHeaders - returns the http headers to return if they are part of the decision
-func (result *EvalResult) GetResponseHTTPHeaders() (http.Header, error) {
```
Pushpalanka, line 163: Hi folks,

This is also **a breaking change** that has went in a minor version release.
Would have been helpful if indicated so in the release note.

I appreciate the effort to simplify the headers approach.

ashutosh-narkar: Thanks for the note @Pushpalanka. We probably missed this while drafting the notes. We'll keep an eye for breaking changes in the future and make sure they're part of the release notes.

Pushpalanka: Thanks for kind consideration @ashutosh-narkar. That will be really helpful.

Pushpalanka: @ashutosh-narkar @regeda 
Are you ok to re-add the existed method option as well?

With the removal of `GetResponseHTTPHeaders() (http.Header, error)` we are facing some inefficiencies(Ref 1, 2) as a library user, having to do an additional conversion make things working.

Another benefit would be, it is more intuitive for a library user to work with `http.Header` from Go, over `*ext_core_v3.HeaderValueOption` specific to envoy.

cc: @mjungsbluth 

[Ref 1](https://github.com/zalando/skipper/pull/3397/files#diff-486bc1f4a1a9405023bb06ca658a7505fefa667b67652b48811ae424b677ca2dR170-R196) 
[Ref 2](https://github.com/zalando/skipper/pull/3397/files#diff-486bc1f4a1a9405023bb06ca658a7505fefa667b67652b48811ae424b677ca2dR170-R196)

ashutosh-narkar:
> Are you ok to re-add the existed method option as well?

I think it should be fine @Pushpalanka. You'll probably re-use the existing logic and export the function, correct?

envoyauth/response.go
regeda: Hi @Pushpalanka 

The additional purpose of this PR is performance optimization where we get rid of an intermediate data in `http.Header`. As far as it's `OPA-to-Envoy` plugin, we convert OPA's data model directly into Envoy's data model.

Unfortunately, there is a breaking change and I believe maintainers must inform the community.

ashutosh-narkar: @regeda with some updates, is it possible to add the `GetResponseHTTPHeaders` method with the original signature?

mjungsbluth: Looking at the changes, I would also think that we can maybe move the actual transformation of the decision to either http.Header or the envoy HeaderOption into a Golang `func(...)` and have the same core logic. This should not have a performance impact and make both the conversations direct and avoid too many allocs...

Pushpalanka: Draft PR added [here](https://github.com/open-policy-agent/opa-envoy-plugin/pull/667/files)

regeda: I think it's possible to wrap the existing function into `GetResponseHTTPHeaders`.

PR #629, opened 2025-01-02
[WIP] Remove intermediate states of the http request conversion to AST

Pushpalanka:
Remove intermediate conversions between the HTTP request to AST input.

Motive: From the below profile capture it was noticed most of the CPU time is spent on these conversions. (It is from a slightly older version of OPA pre v1.0.0)

image

With v1.0.0
image

### DIFF
--- a/envoyauth/request.go
+++ b/envoyauth/request.go
@@ -2,7 +2,6 @@ package envoyauth
 
 import (
 	"encoding/binary"
-	"encoding/json"
 	"fmt"
 	"io"
 	"mime"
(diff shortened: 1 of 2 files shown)


ashutosh-narkar:
Does this [PR](https://github.com/open-policy-agent/opa-envoy-plugin/pull/631) achieve the objective of this PR?

Pushpalanka:
> Does this [PR](https://github.com/open-policy-agent/opa-envoy-plugin/pull/631) achieve the objective of this PR?

Thanks for the reference @ashutosh-narkar.

After having a further look we are now trying the possibility of getting rid of all the intermediate states than just JSON conversions and translating http request directly to the AST.Values. (mainly[ this section](https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/internal/internal.go#L438-L465)) Will update the PR accordingly. Please do share if you see in pitfalls with the approach.

ashutosh-narkar:
That seems reasonable. If you have any benchmarks that show the performance improvement that would be great to see.

PR #631, opened 2025-01-04
refact: use protoreflect to convert check request v3

regeda:
The plugin transforms a `CheckRequest` into OPA's input in two steps:  
1. `protojson.Marshal`  
2. `json.Unmarshal`  

Since `CheckRequest` is a protobuf type, OPA's input can also be generated using protobuf reflection.  

**Benchmark results.**

```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
```

Using JSON as the intermediate format:
```
BenchmarkCheck-12          10000            116460 ns/op           87881 B/op       1734 allocs/op
```

Using protobuf reflection:
```
BenchmarkCheck-12          10000            103498 ns/op           80146 B/op       1592 allocs/op
```

### DIFF
--- a/envoyauth/protomap.go
+++ b/envoyauth/protomap.go
@@ -1,0 +1,65 @@
+package envoyauth
+
+import (
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+type valueResolver func(protoreflect.Value) any
+
+func messageResolver(v protoreflect.Value) any {
+	return protomap(v.Message())
+}
+
+func interfaceResolver(v protoreflect.Value) any {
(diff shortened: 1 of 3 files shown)


srenatus (approved):
Thanks! LGTM, but let's make sure the changelog is clear about this improvement only applying to the v3 version of the API. (Which I assume, by now, most people should be using...?)

regeda:
> Thanks! LGTM, but let's make sure the changelog is clear about this improvement only applying to the v3 version of the API. (Which I assume, by now, most people should be using...?)

I've updated the PR's title. Should I update anything else?

PR #632, opened 2025-01-06
refact: avoid mem alloc with disabled decision log

regeda:
**Benchmark results**.

```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
```

Before:
```
BenchmarkCheck-12             15540             80774 ns/op           55005 B/op       1090 allocs/op
```

After:
```
BenchmarkCheck-12             15696             80721 ns/op           54839 B/op       1087 allocs/op
```

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -34,6 +34,10 @@ type StopFunc = func()
 // TransactionCloser should be called to abort the transaction
 type TransactionCloser func(ctx context.Context, err error) error
 
+func noopTransactionCloser(context.Context, error) error {
+	return nil // no-op default
+}
+
 // NewEvalResult creates a new EvalResult and a StopFunc that is used to stop the timer for metrics
 func NewEvalResult(opts ...func(*EvalResult)) (*EvalResult, StopFunc, error) {
 	var err error
@@ -67,13 +71,9 @@ func NewEvalResult(opts ...func(*EvalResult)) (*EvalResult, StopFunc, error) {
 func (result *EvalResult) GetTxn(ctx context.Context, store storage.Store) (storage.Transaction, TransactionCloser, error) {
 	params := storage.TransactionParams{}
 
-	noopCloser := func(ctx context.Context, err error) error {
-		return nil // no-op default
-	}
-
 	txn, err := store.NewTransaction(ctx, params)
 	if err != nil {
-		return nil, noopCloser, err
+		return nil, noopTransactionCloser, err
 	}
 
 	// Setup a closer function that will abort the transaction.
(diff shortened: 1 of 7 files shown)


envoyauth/response.go
```diff
@@ -67,12 +67,11 @@ func NewEvalResult(opts ...func(*EvalResult)) (*EvalResult, StopFunc, error) {
 func (result *EvalResult) GetTxn(ctx context.Context, store storage.Store) (storage.Transaction, TransactionCloser, error) {
 	params := storage.TransactionParams{}
 
-	noopCloser := func(ctx context.Context, err error) error {
-		return nil // no-op default
-	}
-
 	txn, err := store.NewTransaction(ctx, params)
 	if err != nil {
+		noopCloser := func(ctx context.Context, err error) error {
```
srenatus, line 72: nit
```suggestion
		noopCloser := func(context.Context, error) error {
```

opa/decisionlog/decision_log.go
```diff
@@ -21,15 +20,10 @@ func (e *internalError) Error() string {
 }
 
 // LogDecision - Logs a decision log event
-func LogDecision(ctx context.Context, manager *plugins.Manager, info *server.Info, result *envoyauth.EvalResult, err error) error {
```
srenatus, line 24: Technically this is a breaking change. Someone could be using this function from elsewhere.

The "fix" would be to add another static function here and keep the old one as-is. **But** I think it's _very unlikely_ that someone cares for this function, so let's not be too timid about it. :+1:

srenatus (approved):
Thanks!

envoyauth/response.go
anderseknert: Perhaps this function literal is simple enough that it won't leak to the heap, but you could try to rewrite it a package level function, e.g.

```
func noopCloser(ctx context.Context, err error) error {
    return nil // no-op default
}
```

and then reference that instead. Measure and see if it makes any difference :)

srenatus (approved):
Thank you!

envoyauth/response.go
regeda: I've made this change. However, there are no significant changes from the benchmark perspective.

opa/decisionlog/decision_log.go
```diff
@@ -21,15 +20,10 @@ func (e *internalError) Error() string {
...
-	plugin := logs.Lookup(manager)
-	if plugin == nil {
-		return nil
-	}
-
+func LogDecision(ctx context.Context, plugin *logs.Plugin, info *server.Info, result *envoyauth.EvalResult, err error) error {
```
Pushpalanka, line 23: It's nice improvement!

This is a breaking change with public method signature changing, would be good to be good add in the Change log highlighting the same.

opa/decisionlog/decision_log.go
Pushpalanka: [Skipper](https://github.com/zalando/skipper) is using the function. It would have been nice to have a note in the release note mentioning it's a breaking change. Still since it's 1.0.0 breaking changes were expected.

PR #636, opened 2025-01-07
refact: avoid mem alloc for debug log

regeda:
Each check request creates unnecessary memory garbage for the debug message, even when the log level is set higher than debug.

**Benchmark results.**

```
goos: darwin
goarch: arm64
cpu: Apple M2 Pro
```

With `Debug`:
```
BenchmarkCheck-12            15576             78914 ns/op           54843 B/op       1088 allocs/op
```

Without `Debug`:
```
BenchmarkCheck-12            16891             75411 ns/op           51540 B/op       1037 allocs/op
```

### DIFF
--- a/envoyauth/evaluation.go
+++ b/envoyauth/evaluation.go
@@ -60,11 +60,13 @@ func Eval(ctx context.Context, evalContext EvalContext, input ast.Value, result 
 
 	result.TxnID = result.Txn.ID()
 
-	logger.WithFields(map[string]interface{}{
-		"input": input,
-		"query": evalContext.ParsedQuery().String(),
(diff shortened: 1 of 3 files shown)


anderseknert:
Can't we just wrap that call in a check for log level? The information may be available elsewhere, but if anyone expects to find it there with debug logging turned on, it'd be good if they still found it.

regeda:
> Thanks! Can't we just wrap that call in a check for log level? The information may be available elsewhere, but if anyone expects to find it there with debug logging turned on, it'd be good if they still found it.

Agree. I've wrapped the log statement with a condition.

srenatus (approved):

PR #638, opened 2025-01-09
refact: uuid func, new otelgrpc

regeda:
1. Replace `util.UUID4` function by `google/uuid` package.
2. Replace some deprecation by new versions.

**Benchmark results**

Before:
```
BenchmarkCheck-12             22563             54552 ns/op           34918 B/op        697 allocs/op
```

After:
```
BenchmarkCheck-12             22776             53283 ns/op           34797 B/op        692 allocs/op
```

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -159,8 +159,7 @@ func New(m *plugins.Manager, cfg *Config) plugins.Plugin {
 			otelhttp.WithPropagators(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}, b3.New(b3.WithInjectEncoding(b3.B3MultipleHeader|b3.B3SingleHeader)))),
 		)
 		grpcOpts = append(grpcOpts,
-			grpc.UnaryInterceptor(otelgrpc.UnaryServerInterceptor(grpcTracingOption...)),
-			grpc.StreamInterceptor(otelgrpc.StreamServerInterceptor(grpcTracingOption...)),
+			grpc.StatsHandler(otelgrpc.NewServerHandler(grpcTracingOption...)),
 		)
 	}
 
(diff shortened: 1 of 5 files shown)


srenatus:
Yeah I was surprised, too -- why do we not use one of the common 3rd party deps here? It's not like there weren't at least two in the dependency tree anyways. 😉

regeda:
> Yeah I was surprised, too -- why do we not use one of the common 3rd party deps here? It's not like there weren't at least two in the dependency tree anyways. 😉

I've imported 3rd party library for uuid purposes.

internal/internal.go
```diff
@@ -159,8 +159,7 @@ func New(m *plugins.Manager, cfg *Config) plugins.Plugin {
 			otelhttp.WithPropagators(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}, b3.New(b3.WithInjectEncoding(b3.B3MultipleHeader|b3.B3SingleHeader)))),
 		)
 		grpcOpts = append(grpcOpts,
-			grpc.UnaryInterceptor(otelgrpc.UnaryServerInterceptor(grpcTracingOption...)),
-			grpc.StreamInterceptor(otelgrpc.StreamServerInterceptor(grpcTracingOption...)),
+			grpc.StatsHandler(otelgrpc.NewServerHandler(grpcTracingOption...)),
```
ashutosh-narkar, line 162: Can you please explain this change?

regeda: Both [`otelgrpc.UnaryServerInterceptor`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc#UnaryServerInterceptor) and [`otelgrpc.StreamServerInterceptor`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc#StreamServerInterceptor) are deprecated.

Use [NewServerHandler](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc#NewServerHandler) instead.

ashutosh-narkar (approved):

PR #639, opened 2025-01-10
refact: read bundle revisions if decision log enabled

regeda:
Bundle Revisions are required if the decision log is enabled. 

**Benchmark results**

Before
```
BenchmarkCheck-12            23270             52283 ns/op           34798 B/op        692 allocs/op
```

After
```
BenchmarkCheck-12            24568             49867 ns/op           34316 B/op        677 allocs/op
```

### DIFF
--- a/envoyauth/evaluation.go
+++ b/envoyauth/evaluation.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	"github.com/open-policy-agent/opa/ast"
-	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/config"
 	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/rego"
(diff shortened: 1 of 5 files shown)


envoyauth/evaluation.go
```diff
@@ -32,8 +31,7 @@ type EvalContext interface {
 
 // PrepareQueryOpts - Options to prepare a Rego query to be passed to the CreatePreparedQueryOnce method
 type PrepareQueryOpts struct {
-	Opts        []func(*rego.Rego)
-	PrepareOpts []rego.PrepareOption
+	Opts []func(*rego.Rego)
```
ashutosh-narkar, line 34: This would be a breaking change. I'm not quite sure why we need to remove this and what the purpose of the change is. It would be helpful to understand this. Thanks!

regeda: There is no usage of this field. However, I could revert this change if it has concerns.

ashutosh-narkar: Let's revert the change in this PR as it's unrelated and also backwards-incompatible. Thanks.

regeda: PTAL

ashutosh-narkar (approved):

PR #663, opened 2025-02-06
feat: support query_parameters_to_set in response

sedovmik:
Add support for query_parameters_to_set field in response to allow setting/overwriting query parameters before sending upstream. Similar to existing query_parameters_to_remove support.

Example usage in OPA policies:

```rego
package envoy.authz

default allow = true

query_parameters_to_set = {
   "user_id": "123",
   "tenant": "acme"
}

query_parameters_to_set = {
   "role": ["admin", "user"],
   "scope": ["read", "write"]
}

query_parameters_to_set = {
   "user_id": "123",
   "role": ["admin", "user"]
}

result["allowed"] = allow
result["query_parameters_to_set"] = query_parameters_to_set

```

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -441,3 +441,51 @@ func preallocateForHTTPHeaders(result *http.Header, _ int) {
 		*result = make(http.Header)
 	}
 }
+
+// GetRequestQueryParametersToSet returns the query parameters to set in the request
+func (result *EvalResult) GetRequestQueryParametersToSet() ([]*ext_core_v3.QueryParameter, error) {
+	switch decision := result.Decision.(type) {
+	case bool:
+		return nil, nil
+	case map[string]interface{}:
+		val, ok := decision["query_parameters_to_set"]
+		if !ok {
+			return nil, nil
+		}
+
+		params, ok := val.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("type assertion error, expected query_parameters_to_set to be a map but got '%T'", val)
+		}
+
+		result := make([]*ext_core_v3.QueryParameter, 0, len(params))
+
+		for key, value := range params {
+			switch v := value.(type) {
+			case string:
+				result = append(result, &ext_core_v3.QueryParameter{
+					Key:   key,
+					Value: v,
+				})
+			case []interface{}:
+				result = slices.Grow(result, len(v))
+				for _, item := range v {
+					strItem, ok := item.(string)
+					if !ok {
+						return nil, fmt.Errorf("type assertion error: expected array element to be string but got '%T'", item)
+					}
+					result = append(result, &ext_core_v3.QueryParameter{
+						Key:   key,
+						Value: strItem,
+					})
+				}
+			default:
+				return nil, fmt.Errorf("type assertion error, expected value to be string or array but got '%T'", value)
+			}
+		}
+
+		return result, nil
+	}
+
+	return nil, result.invalidDecisionErr()
+}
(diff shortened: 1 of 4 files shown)


internal/internal_test.go
```diff
@@ -1562,6 +1562,64 @@ func TestCheckAllowObjectDecisionReqQueryParamsToRemove(t *testing.T) {
...
+	if err := util.Unmarshal([]byte(exampleAllowedRequest), &req); err != nil {
+		panic(err)
+	}
+
+	module := `
+		package envoy.authz
+
+		default allow = true
+
+		query_parameters_to_set := [
```
regeda, line 1576: Great contribution!

Could you consider the following API:
```
query_parameters_to_set := {
  "foo": "value1",
  "bar": "value2",
}
```

It leads to less iterations and less memory allocations.

sedovmik: Thank you for the comment! I agree that the format could be simplified. I intended to keep it aligned with Envoy's config.core.v3.QueryParameter for implementation simplicity, but your suggestion is indeed more ergonomic.

Regarding multi-value parameters - In HTTP, it's possible to have multiple values for the same query parameter (e.g., ?foo=value1&foo=value2). To properly support this, I could add support for both value types (string and array of strings):

```rego
query_parameters_to_set := {
  "foo": "value1",             # single value
  "bar": ["value1", "value2"]  # multiple values
}
```

What are your thoughts? I'd be happy to update the implementation accordingly.

regeda: I fully support both value types.

envoyauth/response.go
```diff
@@ -394,3 +394,50 @@ func transformHeadersToEnvoy(input any) ([]*ext_core_v3.HeaderValueOption, error
...
+		if !ok {
+			return nil, nil
+		}
+
+		params, ok := val.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("type assertion error, expected query_parameters_to_set to be a map but got '%T'", val)
+		}
+
+		var result []*ext_core_v3.QueryParameter
```
regeda, line 414: Could you preallocate the capacity to improve the performance?
```suggestion
		result := make([]*ext_core_v3.QueryParameter, 0, len(params))
```

envoyauth/response.go
```diff
@@ -394,3 +394,50 @@ func transformHeadersToEnvoy(input any) ([]*ext_core_v3.HeaderValueOption, error
...
+
+		for key, value := range params {
+			switch v := value.(type) {
+			case string:
+				result = append(result, &ext_core_v3.QueryParameter{
+					Key:   key,
+					Value: v,
+				})
+			case []interface{}:
+				for _, item := range v {
```
regeda, line 424: Here you also could grow the slice using [`slices.Grow`](https://pkg.go.dev/slices#Grow):
```suggestion
				result = slices.Grow(result, len(v))
				for _, item := range v {
```

regeda (changes requested):

regeda:
Please add this feature to the official OPA documentation (like https://github.com/open-policy-agent/opa/pull/7241)

sedovmik:
@regeda Thank you for the review. The documentation PR is available here: https://github.com/open-policy-agent/opa/pull/7405

regeda (approved):

sedovmik:
@tjons @ashutosh-narkar 

I rework the change with all @regeda's suggestions, including the docs update in the main repo. Everything is ready for another look when you get a chance.

sedovmik:
> You'll need to update [this](https://www.openpolicyagent.org/docs/latest/envoy-primer/) page.

Yes, PR for documentation is ready: https://github.com/open-policy-agent/opa/pull/7405

ashutosh-narkar (approved):

PR #665, opened 2025-02-07
fix: build badge url

regeda:
Set a valid URL to the Build Badge

Before:
[Build Status](https://github.com/open-policy-agent/opa-envoy-plugin/actions)

After:
[Build Status](https://github.com/open-policy-agent/opa-envoy-plugin/actions)


srenatus (approved):
Thank you!

PR #667, opened 2025-02-11
Re-introduce GetResponseHTTPHeaders method

Pushpalanka:
**Context:**
Re-introduce the removed method from https://github.com/open-policy-agent/opa-envoy-plugin/pull/628 using separate transform functions following [this comment](https://github.com/open-policy-agent/opa-envoy-plugin/pull/628/files#r1948644915) from @mjungsbluth.

Both signatures will be available to use as required,
```
func (result *EvalResult) GetResponseEnvoyHeaderValueOptions() ([]*ext_core_v3.HeaderValueOption, error) {
func (result *EvalResult) GetResponseHTTPHeaders() (http.Header, error) {
```


envoyauth/response.go
```diff
@@ -394,3 +412,54 @@ func transformHeadersToEnvoy(input any) ([]*ext_core_v3.HeaderValueOption, error
 	}
 	return nil, fmt.Errorf("type assertion error, expected headers to be of type 'object' but got '%T'", input)
 }
+
+func transformToHTTPHeaderFormat(input interface{}) (http.Header, error) {
+	result := make(http.Header)
+
+	takeResponseHeaders := func(headers map[string]interface{}) error {
+		for key, value := range headers {
+			switch values := value.(type) {
```
mjungsbluth, line 421: This seems to be a structural duplicate to the transformation for the envoy HttpOption. Do you want to unify this more?

Pushpalanka: Yes indeed. Addressed.

mjungsbluth: What I meant is that this switch statement in the for loop is essentially the same as in `makeEnvoyHeaderValueOptionsFromHeadersMap` so you could model this with a transformer + collector function and then the same for loop and structural swich could be used...

Pushpalanka: Thanks for clarifying. I hope it is there now.

envoyauth/response.go
```diff
@@ -343,54 +330,100 @@ func makeHeaderValueOption(k, v string) *ext_core_v3.HeaderValueOption {
...
+	collector func(string, string, *T),
+	result *T,
+) error {
 	for key, value := range headers {
 		switch val := value.(type) {
 		case string:
-			hvo = append(hvo, makeHeaderValueOption(key, val))
+			collector(key, val, result)
 		case []string:
-			hvo = slices.Grow(hvo, len(val))
```
mjungsbluth, line 353: Removing this might yield a small performance hit as the array will need to grow for each sub-element in the Envoy case.

envoyauth/response.go
```diff
@@ -422,8 +428,15 @@ func collectEnvoyHeaders(key, value string, result *[]*ext_core_v3.HeaderValueOp
...
+}
+
+func preallocateForEnvoyHeaders(result *[]*ext_core_v3.HeaderValueOption, additional int) {
+	*result = slices.Grow(*result, additional)
+}
+
+func preallocateForHTTPHeaders(result *http.Header, additional int) {
 	if *result == nil {
-		*result = make(http.Header)
+		*result = make(http.Header, additional)
```
mjungsbluth, line 440: I think this grows the map itself and not the values for a specific key.
Maybe that is also not easily possiblr

Pushpalanka: Yes, that's true. Will ignore the additional value, but keep for the function signature to match.

envoyauth/response_test.go
```diff
@@ -455,6 +455,127 @@ func TestGetResponseHeaders(t *testing.T) {
 	if len(result) != 2 {
 		t.Fatalf("Expected two header but got %v", len(result))
 	}
+	//
```
mjungsbluth, line 458: Commented code to be removed?

mjungsbluth:
LGTM in general!

envoyauth/response.go
```diff
@@ -343,54 +331,112 @@ func makeHeaderValueOption(k, v string) *ext_core_v3.HeaderValueOption {
...
+
+func extractHeadersFromDecision(decision map[string]interface{}, fieldName string) ([]map[string]interface{}, error) {
+	val, ok := decision[fieldName]
+	if !ok {
+		return nil, nil
+	}
+
+	switch v := val.(type) {
+	case map[string]interface{}:
+		return []map[string]interface{}{v}, nil
```
regeda, line 378: Please, avoid additional allocations. We have a highload service based on OPA where each allocation drastically impacts the performance.

envoyauth/response.go
```diff
@@ -343,54 +331,112 @@ func makeHeaderValueOption(k, v string) *ext_core_v3.HeaderValueOption {
...
-	hvo = slices.Grow(hvo, len(headers))
+func (result *EvalResult) getHeadersFromDecision(fieldName string) ([]*ext_core_v3.HeaderValueOption, error) {
+	return getHeadersWithTransformation(result.Decision, fieldName, preallocateForEnvoyHeaders, collectEnvoyHeaders)
+}
+
+func (result *EvalResult) getHTTPHeadersFromDecision(fieldName string) (http.Header, error) {
+	return getHeadersWithTransformation(result.Decision, fieldName, preallocateForHTTPHeaders, collectHTTPHeaders)
+}
+
+func getHeadersWithTransformation[T any](
```
regeda, line 342: IMHO, a new approach looks a bit complicated and not easy to follow the logic.

Could you make just a function which converts the result from `getHeadersFromDecision` into `http.Header`?

regeda (changes requested):

envoyauth/response.go
regeda:
```suggestion
func (result *EvalResult) getHTTPHeadersFromDecision(fieldName string) (http.Header, error) {
	opts, err := result.getHeadersFromDecision(fieldName)
	if err != nil {
		return nil, err
	}
	header := make(http.Header, len(opts))
	for _, opt = range opts {
		header.Add(opt.GetHeader().GetKey(), opt.GetHeader().GetValue())
	}
	return header, nil
}
```

regeda (changes requested):

envoyauth/response_test.go
Pushpalanka: Mistake, will uncomment it. Introduced it to valid the case of `[]interface{}` for header values.

envoyauth/response.go
Pushpalanka: It's is an effort to balance complexity and performance for both cases of returning `HeaderValueOption` and `http.Header`. 
This suggested approach introduces an unnecessary intermediate conversion to `ext_core_v3.HeaderValueOption` for http.Header returning case affecting performance in that scenario. 
This concern directs us towards the implemented approach which keep both the cases optimized for highload services.

regeda: Could you introduce the benchmark of `getHeadersFromDecision` before and after your refactoring?

regeda: Moreover, you already made an intermediate conversion using `extractHeadersFromDecision` function.

regeda (changes requested):

envoyauth/response.go
anderseknert: @regeda I respect your requirements, but it's probably better to follow up with PRs of your own to improve performance where needed rather than making requests for that in others contributions.

envoyauth/response.go
```diff
@@ -4,9 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"net/http"
-	"slices"
```
srenatus, line 8: I kind of appreciated the blocks here -- stdlib vs third-party. but that's not a blocker.

anderseknert (approved):

srenatus (approved):
Thanks! LGTM 🎈

envoyauth/response.go
regeda: I'm kind of responsible for this breaking change.

I would help to set it back. PTAL https://github.com/open-policy-agent/opa-envoy-plugin/pull/668

anderseknert: That's fair! Let's have this merged and follow up with additions based off of this.

anderseknert: Also, if you're interested in reducing allocations, you'll find a lot to like in OPA v1.2.0 🙂

PR #668, opened 2025-02-19
feat: return headers as http.Header

regeda:
The PR https://github.com/open-policy-agent/opa-envoy-plugin/pull/628 contains a breaking change. The following function has been removed:
```
func (result *EvalResult) GetResponseHTTPHeaders() (http.Header, error)
```

I would reintroduce it keeping the actual state of logic.

**Benchmark**.

_Before_
```
BenchmarkCheck-12           17308             70453 ns/op           47346 B/op        964 allocs/op
```

_After_
```
BenchmarkCheck-12           17799             69564 ns/op           47317 B/op        960 allocs/op
```

### DIFF
--- a/envoyauth/response.go
+++ b/envoyauth/response.go
@@ -185,19 +185,72 @@ func (result *EvalResult) GetRequestHTTPHeadersToRemove() ([]string, error) {
 	return result.getStringSliceFromDecision("request_headers_to_remove")
 }
 
+func (result *EvalResult) makeHeadersFromDecision(fieldName string, hb headersBag) error {
+	switch decision := result.Decision.(type) {
+	case bool:
+		return nil
+	case map[string]any:
+		val, ok := decision[fieldName]
+		if !ok {
+			return nil
+		}
+
+		return transformAnyInputToHeadersBag(val, hb)
+	default:
+		return result.invalidDecisionErr()
+	}
+}
+
+type headersBag interface {
+	allocate(n int)
+	add(k, v string)
+}
+
+type envoyHeaders []*ext_core_v3.HeaderValueOption
+
+func (h *envoyHeaders) add(k, v string) {
+	*h = append(*h, &ext_core_v3.HeaderValueOption{
+		Header: &ext_core_v3.HeaderValue{
+			Key:   k,
+			Value: v,
+		},
+	})
+}
+
+func (h *envoyHeaders) allocate(n int) {
+	*h = slices.Grow(*h, n)
+}
+
+type httpHeaders struct {
+	http.Header
+}
+
+func (h *httpHeaders) add(k, v string) {
+	h.Add(k, v)
+}
+
+func (h *httpHeaders) allocate(n int) {
+	if h.Header == nil {
+		h.Header = make(http.Header, n)
+	}
+}
+
 // GetResponseHTTPHeaders - returns the http headers to return if they are part of the decision as http header
 func (result *EvalResult) GetResponseHTTPHeaders() (http.Header, error) {
-	return result.getHTTPHeadersFromDecision("headers")
+	var h httpHeaders
+	return h.Header, result.makeHeadersFromDecision("headers", &h)
 }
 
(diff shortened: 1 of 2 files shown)


envoyauth/response.go
```diff
@@ -185,30 +185,67 @@ func (result *EvalResult) GetRequestHTTPHeadersToRemove() ([]string, error) {
...
+	*h = append(*h, makeHeaderValueOption(k, v))
+}
+
+func (h *envoyHeaders) allocate(n int) {
+	*h = slices.Grow(*h, n)
+}
+
+type httpHeaders struct {
+	http.Header
+}
```
srenatus, line 221: Do you need the `struct {}`? Wouldn't one of these do the trick, too? Frankly, I cannot remember which one. 😅 
```go
type httpHeaders = http.Header

type httpHeaders http.Header
```

regeda: Yes, `http.Header` should be wrapped into a struct.

The following variants don't allow adding new functions.

envoyauth/response.go
```diff
@@ -184,19 +185,72 @@ func (result *EvalResult) GetRequestHTTPHeadersToRemove() ([]string, error) {
...
+			return nil
+		}
+
+		return transformAnyInputToHeadersBaggage(val, hb)
+	default:
+		return result.invalidDecisionErr()
+	}
+}
+
+type headersBaggage interface {
```
tjons, line 204:
```suggestion
type headersContainer interface {
```

I suggest that we name this something other than `headersBaggage`. `headersContainer`, `headersBag`, etc all are fine - baggage (in native english) implies that this is unimportant/unwanted.

envoyauth/response.go
```diff
@@ -184,19 +185,72 @@ func (result *EvalResult) GetRequestHTTPHeadersToRemove() ([]string, error) {
 	return result.getStringSliceFromDecision("request_headers_to_remove")
 }
 
+func (result *EvalResult) makeHeadersFromDecision(fieldName string, hb headersBaggage) error {
+	switch decision := result.Decision.(type) {
+	case bool:
+		return nil
+	case map[string]interface{}:
```
tjons, line 192:
```suggestion
	case map[string]any:
```

tjons (commented):
Couple nits, generally looks good! Thanks for adding this back in

Pushpalanka (commented):
The method is already re-added in this PR - https://github.com/open-policy-agent/opa-envoy-plugin/pull/667

Performance has been degraded with this, based on the shared benchmark results,
- execution time is ↓ Slower by ~1.8%
- ↓ fewer operations (~1.2% decrease)
- memory usage ↑ Increased by 2 bytes (negligible change)
- allocation no change.

So what is the objective or what is the gain we are looking forward in return?

regeda:
> The method is already re-added in this PR - #667
> 
> Performance has been degraded with this, based on the shared benchmark results,
> 
>     * execution time is ↓ Slower by ~1.8%
> 
>     * ↓ fewer operations (~1.2% decrease)
> 
>     * memory usage ↑ Increased by 2 bytes (negligible change)
> 
>     * allocation no change.
> 
> 
> So what is the objective or what is the gain we are looking forward in return?

I've added headers reading to the benchmark. We could see the consumption with some result: `headers`, `headers_to_remove`, `query_parameters_to_remove`, `response_headers_to_add`.

regeda:
Hi @ashutosh-narkar

This PR is hanging for a while. Could you review it and merge? There are some improvements in the benchmark test and plugin's performance.

tjons:
@regeda I'd like to help you get this merged - sorry for the wait here! I inadvertently `merged` main instead of `rebasing` it, and the commit isn't showing up for me locally. Could you rebase `main` onto your branch and force-push and then request my review? I'll approve.

regeda:
> @regeda I'd like to help you get this merged - sorry for the wait here! I inadvertently `merged` main instead of `rebasing` it, and the commit isn't showing up for me locally. Could you rebase `main` onto your branch and force-push and then request my review? I'll approve.

I've rebased my branch.

tjons (approved):

PR #676, opened 2025-03-06
perf: use ast.Value and []string in request input

regeda:
**Before**:
```
BenchmarkCheck-12          28180             42507 ns/op           28437 B/op        570 allocs/op
```

After:
```
BenchmarkCheck-12          29908             40675 ns/op           27856 B/op        552 allocs/op
```


tjons (approved):
LGTM - nice optimization!

PR #689, opened 2025-04-05
Update envoy proxy versions used in tests

tjons:
Our tests currently use Envoy 1.10 - an _ancient version_ of Envoy. This PR updates the tests to use 1.33 to make sure that we continue to support recent versions of Envoy, as well as to provide users with configuration examples that are pertinent to versions currently in use.


anderseknert (approved):
Nice!

PR #702, opened 2025-05-02
Bump OPA to 1.4.2

sspaink:


tjons (approved):
@srenatus @johanfylling perfect example of how removing `vendor/` would be a quality of life improvement :)

PR #715, opened 2025-05-27
docs: Correct OPA links

charlieegan3:
x-ref https://github.com/open-policy-agent/opa/pull/7629

Also fix some links to point to the correct locations.


johanfylling (approved):
👍

PR #738, opened 2025-06-28
reenable govet linter and fix its violations

### LINKED ISSUE #727: Enable the `govet` linter
The `govet` linter has been disabled because when switching to `golangci-lint`, there were too many errors. Reenable the `govet` linter and fix the issues.

akashsawan1:
- Replace `err != tc.err` with `errors.Is(err, tc.err)` in envoyauth/request_test.go
- Swap `reflect.DeepEqual` for `==` in internal/internal_test.go
- Use package-level variable ErrInvalidPath for consistent errors.Is() checks

Resolves #727

### DIFF
--- a/envoyauth/request.go
+++ b/envoyauth/request.go
@@ -3,6 +3,7 @@ package envoyauth
 import (
 	"encoding/binary"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"mime"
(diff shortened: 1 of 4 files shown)


envoyauth/request.go
```diff
@@ -32,6 +33,7 @@ var v3Info = ast.NewObject(
 	keyValue("ext_authz", "v3"),
 	keyValue("encoding", "protojson"),
 )
+var ErrInvalidPath = errors.New("invalid parsed path")
```
tjons, line 36: Do you think we need to export this? I think we can rename it `errInvalidPath` and avoid making it part of the package's interface.

tjons (commented):
One comment, otherwise LGTM! Thanks for yet another great contribution. Can you rebase `main` on this as well?

tjons (approved):

PR #742, opened 2025-06-30
Bump OPA to 1.6.0

johanfylling:


tjons (approved):
LGTM! Could we hold the 1.6.0 release of the plugin until we can merge https://github.com/open-policy-agent/opa-envoy-plugin/pull/733 and https://github.com/open-policy-agent/opa-envoy-plugin/pull/732 too? Would be nice to get them out

PR #748, opened 2025-07-09
value-cache-added-to-evalcontext

amak3:
To be able to use JWT validation cache feature enabled in 1.1.0 OPA (see [changelog](https://github.com/open-policy-agent/opa/releases/tag/v1.1.0) InterQueryBuiltinValueCache needs to be added to EvalContext so it can be used as EvalOption. This way it will be later propagated in BuiltinContext in OPA.


srenatus (dismissed):
Thank you, that's a nice contribution. Good find, this should really be passed along 👍

PR #754, opened 2025-07-30
Feature/iam 481 xml dc plugin

WojciechLuczkow-Flyr:


tjons (changes requested):
Hi, could you tell us a little more about this PR and what you're trying to accomplish?

WojciechLuczkow-Flyr:
Hello @tjons Sorry for that, I have created that PR by mistake

srenatus:
You know, you don't need to fork opa-envoy-plugin to add a builtin. Just create your own project that uses something like [this main.go](https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/cmd/opa-envoy-plugin/main.go) (without any internal imports!), and adds the builtins.

PR #773, opened 2025-09-28
Fix/check context cancel logging

abdullahalrifat:
This PR adds a unit test to verify that Check correctly handles explicit context cancellation before query execution.

- Introduced a testServer wrapper with a beforeCheck hook to simulate cancellation during request handling.

- Ensured that the expected error (context canceled) is returned and properly logged.

- Validated that the error metric CheckRequestCancelledErr is incremented.

This strengthens test coverage around graceful handling of cancelled requests and ensures consistent logging/metrics behavior.


internal/internal.go
```diff
@@ -461,9 +461,9 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req any) (*ext_auth
 		}, stop, nil
 	}
 
-	if ctx.Err() != nil {
-		err = errors.Wrap(ctx.Err(), "check request timed out before query execution")
-		internalErr = newInternalError(CheckRequestTimeoutErr, err)
+	if thrownError, ok := checkCtxError(ctx); ok {
```
tjons, line 464: Go doesn't really have the notion of "thrown errors", errors are values and are not handled as exceptions. Could you name this something like `ctxErr`

internal/internal.go
```diff
@@ -758,3 +758,21 @@ func v2Status(s *ext_type_v3.HttpStatus) *ext_type_v2.HttpStatus {
 		Code: ext_type_v2.StatusCode(s.Code),
 	}
 }
+
+// checkCtxError checks if the context has error and returns appropriate internal error
+func checkCtxError(ctx context.Context) (*Error, bool) {
```
tjons, line 763: I believe this function can be simplified by just returning a `*Error` type and omitting the `bool` return value, since the `bool` is always true when the `*Error` is not nil, but false when the `*Error` is nil. Is there some other information you were hoping to use this `bool` to represent?

internal/internal.go
```diff
@@ -758,3 +758,21 @@ func v2Status(s *ext_type_v3.HttpStatus) *ext_type_v2.HttpStatus {
 		Code: ext_type_v2.StatusCode(s.Code),
 	}
 }
+
+// checkCtxError checks if the context has error and returns appropriate internal error
```
tjons, line 762:
```suggestion
// checkCtxError checks if the context has an error and returns an appropriate internal Error
```

internal/internal_test.go
```diff
@@ -622,6 +622,65 @@ func TestCheckContextTimeout(t *testing.T) {
...
+	if s.beforeCheck != nil {
+		s.beforeCheck()
+	}
+	return s.envoyExtAuthzGrpcServer.Check(ctx, req)
+}
+
+func TestCheckContextCancelled(t *testing.T) {
+	var req ext_authz.CheckRequest
+	if err := util.Unmarshal([]byte(exampleAllowedRequest), &req); err != nil {
+		panic(err)
```
tjons, line 640: We should use `t.Error` here to convey an error message for the test failure, rather than just panic. I'm not sure you need to stop execution of other tests anyways, which `panic()/t.Fail()` will do.

tjons (changes requested):
Hi @abdullahalrifat - thanks for the contribution. I left a review with some suggested changes, but I'd be happy to help you get this in. That said, I'm curious how you found this behavior? Did it arise in a production deployment? Some background about how this context might be cancelled would be helpful - are you seeing context cancellations from Envoy?

Once we better understand that and you are able to make the suggested changes, I will need you to squash your commits and rebase on top of the latest `main` as well.

Thanks for contributing!

internal/internal.go
abdullahalrifat: fixed

internal/internal_test.go
abdullahalrifat: i just kept it similar to other tests 
```
if err := util.Unmarshal([]byte(exampleDeniedRequest), &req); err != nil {
		panic(err)
	}
```

tjons (commented):
This LGTM, @abdullahalrifat - can you please rebase on the latest `main` and squash your commits?

PR #778, opened 2025-10-31
bump opa to v1.10.0

sspaink:
To accommodate the new WASM bump deleting the vendor directory was needed for the CI to pass.
Thanks @srenatus for helping me with this!

___ 

Tried to use Dependabot for this, but while it found the latest OPA version and the Github Action succeeds the PR isn't created and doesn't show a specific error:

Screenshot 2025-10-31 at 11 23 03 AM


srenatus (approved):
💣

PR #876, opened 2026-07-31
internal: decouple decision logging from request context cancellation

### LINKED ISSUE #647: Log decisions when context is already cancelled
Currently the opa-envoy plugins logs decisions in all conditions (ie. presence/absence of errors). For example, when the context is cancelled or deadline is reached before query execution, the plugin passes the same context to the decision logger routine in order to log the decision. The decision log plugin then tries to [evaluate the log masking policy](https://github.com/open-policy-agent/opa/blob/main/plugins/logs/plugin.go#L849) and passes it the same context. The `rego.eval` method eventually gets called and it contains a [custom query cancellation mechanism](https://github.com/open-policy-agent/opa/blob/main/rego/rego.go#L2382) which leads to a race between the context cancellation and  query evaluation completion. As a result in some cases the decision event does not get logged. This scenario can be reproduced by running [this](https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/internal/internal_test.go#L424) test multiple times.

Something that needs to be explored here is whether a decision event should even be logged when the context is cancelled or deadline is reached before query execution. One could argue that users may expect to see a corresponding decision log event.

pujitha24:
Motivation:

When a check request's context is already cancelled or its deadline has
passed before query evaluation runs, the decision is still logged, but
`logDecision` was called with that same cancelled/expired context. OPA's
decision log plugin evaluates a mask policy using `rego.Eval` with the
context it is given; if the context is already done, a cancellation
goroutine races the eval and can cause it to abort early. When that
happens, the decision log plugin logs an internal error and drops the
decision event instead of emitting it. This is a race, not deterministic:
it depends on timing, so it does not reproduce on every cancelled/timed
out request, and it does not crash the server or affect the response
returned to Envoy — only the decision log event for that request can be
silently lost.

Approach:

In the `stop` closure of `check()`, log the decision using
`context.WithoutCancel(ctx)` instead of `ctx`. This preserves context
values (e.g. trace span info used later via `trace.SpanFromContext`) but
removes the cancellation signal and deadline, so the mask policy
evaluation (and any other context-aware work in `logDecision`, e.g.
`ReadRevisions`) can no longer race against an already-cancelled/expired
context. This matches the behavior requested in the issue: decisions
should be logged regardless of whether the request context was cancelled
before query execution.

Validation:

- go build ./...
- go test ./internal/... (all passing, including new regression test
  TestCheckContextCancelledLogsWithUncancelledContext, which cancels the
  request context before query evaluation and asserts, via a context-
  capturing decision-log plugin stub, that the context handed to the
  logger is not cancelled; confirmed this test fails with "context
  canceled" on the pre-fix code and passes after the fix)

Fixes #647

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>


internal/internal.go
```diff
@@ -425,7 +425,12 @@ func (p *envoyExtAuthzGrpcServer) check(ctx context.Context, req any) (*ext_auth
 				p.metricErrorCounter.With(prometheus.Labels{"reason": internalErr.Code}).Inc()
 			}
 		}
-		logErr := p.logDecision(ctx, input, result, err)
+		// Log the decision using a context that is no longer subject to the
+		// original request's cancellation/deadline. The check request may
+		// have been cancelled or timed out before query execution, but the
+		// decision should still be logged, e.g. by evaluating a mask policy,
+		// without racing against a context that is already done.
+		logErr := p.logDecision(context.WithoutCancel(ctx), input, result, err)
```
srenatus, line 433: TIL: `context.WithoutCancel(ctx)` 💡

srenatus (approved):

srenatus:
Thank you! I went ahead and fixed the same issue over in OPA.

